PR #15, opened 2021-02-01
Add better configuration management

algae12:
- Removed configman() and replaced with configparser() based management.
- Updated README and added a requirments.txt file.


edeng23:
Looking Good 🏅 
Much appreciated!

PR #41, opened 2021-02-12
Add a Debug in screen.

electrotelco:
Fixed APIError(code=-2013): Order does not exist.  + debug and others...

Added a collection of print's to observe the operation of the bot, in this way, you can see the information on the screen, not having to access the log.

Similarly, it has been added to ask you for the initial currency.
The print's are made in Spanish, for another language, that each one adjusts to their preference.

In development / ideas to implement:
-> Assignable loss limit (per input) or fixed up to 3%
-> When selling, test the coin value, if it is rising, continue in the test cycle, to maximize the selling price (and profit).
-> Being able to start from USDT, indicating an amount to use, in this way, it will not use the 100% that you have in USDT, but what you indicate.

Development team: Electrotelco, JuanXo304, TurboCoin.
Credits / Initial Base: https://github.com/edeng23/binance-trade-bot

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -147,12 +147,16 @@ def buy_alt(client, alt_symbol, crypto_symbol):
             time.sleep(10)
 
     while stat[u'status'] != 'FILLED':
-        stat = client.get_order(
-            symbol=alt_symbol+crypto_symbol, orderId=order[u'orderId'])
-        time.sleep(1)
+       try:
+            symbol=alt_symbol+crypto_symbol, orderId=order[u'orderId'])	            stat = client.get_order(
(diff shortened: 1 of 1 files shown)


thiazzz:
Why not do it like this:

ch = logging.StreamHandler()
ch.setLevel(logging.DEBUG)
ch.setFormatter(formatter)
logger.addHandler(ch)

edeng23:
From what I see, all you did was add a single print which is in spanish... 
I agree with @thiazzz on the standard way to achieve the debug prints.

electrotelco:
I'm a newbie to Python and I'm learning as I go.

I added print lines in the parts that I consider important to see in debug (like adding timestamp to them).

It would be great to add it to the main one!

edeng23:
The logs are now also printed.

PR #42, opened 2021-02-12
Telegram logger

thiazzz:
Added telegram logger as per this article: https://www.marcodena.it/blog/telegram-logging-handler-for-python-java-bash/

Also added a logger to the console


edeng23:
The bot crashes with a timeout error (telegram.error.TimedOut: Timed out) during the trade threshold initialisation stage. Does it work for you?
Edit: Nevermind, I just realised that I ran the previous version of your code. With the new version I see that if a Telegram message fails, the bot just continues to run without sending it. That's fine in my opinion since it will still appear in the logfile.
I'll run some more tests and merge the code :)

edeng23:
Works well for me, thanks for the generous contribution.

PR #67, opened 2021-02-17
Rename requirments.txt to requirements.txt

darmiel:
Hi 👋,

this is a very small change, but during setup I almost went crazy because I kept typing "requirements.txt" but the file was not found because the filename was misspelled here 😄.


edeng23:
Fair enough :)

PR #73, opened 2021-02-17
New transaction logic

xjakubo:
I changed transaction logic, changed string formatting style in these (using fstrings)

-implemented roundDown function to use all of available funds, simply cutting decimal places to stepSize of coin (ignored small coins).
-moved order to function for better visibility.
-added order_quantity to try function, so if coin blows up it adjusts price.
-added try-catch to all network functions.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -77,8 +77,24 @@ if TELEGRAM_TOKEN:
     logger.addHandler(th)
 
 logger.info('Started')
-
-supported_coin_list = []
+#you need to add decimal places down here, so if you need to buy TRX binance accepts 1 decimal place (0,1 = 10) ICX (0,01 = 100) etc
+larger_coins_digits = {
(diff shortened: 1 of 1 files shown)


crypto_trading.py
```diff
@@ -77,8 +77,24 @@ def format(self, record):
...
+'TOMO': 100,
+'EOS': 100,
+'QTUM': 1000,
+'ETC': 100,
+'XMR': 100000,
+'DASH': 100000,
+'NEO': 1000,
+'ATOM': 1000,
+'OMG': 100
+}
```
german-bortoli, line 97: Why is this needed? Can't be something more dynamic than hardcoding it ?

edeng23:
@xjakubo Are you still sure that this is necessary and not part of the code already?

rel-s:
This PR will be closed for now until it is explaind & rebased.

PR #76, opened 2021-02-18
Initialize, update and scout ticker prices in batch - Ignore symbols in supported_coins_table not matching remote pair - Telegram log with QueueHandler (no blocking)

tinchogon34:
I use get_all_tickers function to retrieve all tickers in one request and then find the ticker by symbol there.

I had to change Telegram logger to QueueHandler so it don't block until request ends when there are too many requests, in initialize_trade_thresholds for example.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -2,6 +2,7 @@
 from binance.client import Client
 from binance.exceptions import BinanceAPIException
 import logging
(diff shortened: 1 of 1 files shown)


edeng23:
Great! Pulled #79

tinchogon34:
Great!

PR #77, opened 2021-02-18
Merging Supported Coin List into user.cfg

rel-s:
Deleted the old `supported_coin_list` file, and moved it into a new section in `user.cfg`. Now changing this file shouldn't cause any tracked events.


RawUnited (commented):
21 user.cfg.example works fine.
15  crypto_trading.py breaks early needs improvement.

rel-s:
What exactly broke in your test?

PR #84, opened 2021-02-19
Direct Pair Transactions

CTetford:
- Retrieves all available tickers at startup as a reference of what currency pairs exist.
- At each transaction attempt it will check if a given pair exists and performs a direct trade. If it doesn't exist, the transaction then occurs via the bridge currency.
- This will both reduce transaction fees and speed up transaction times.

Note: I haven't fully tested this yet since my system seems to be stuck on one currency at the moment, but I thought others may want to test it.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -322,16 +322,28 @@ def sell_alt(client, alt_symbol, crypto_symbol):
     return order
 
 
-def transaction_through_tether(client, source_coin, dest_coin):
+def transaction(client, source_coin, dest_coin, available_tickers):
     '''
-    Jump from the source coin to the destination coin through tether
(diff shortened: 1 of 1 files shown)


ghost:
I tried the script. The supported coins doc only listed eth and BTC. No exchanges were some after 8 hours or so and after 6 hours I got an error:

Error while scouting
Read timeout

CTetford:
The catch-all try/except in the main loop is frustratingly annoying - it makes it impossible to figure out what the actual issue is. 

I have no idea what caused your error. Can you disable it so it actually outputs the line the error occurs on and try again?

CTetford:
I need to edit this to use a limit buy for both scenarios of direct pairs. Currently its only on one of them. 

Avoid using it for now.

CTetford:
Closing this. I'll re-open when I've worked on it some more.

PR #85, opened 2021-02-19
Fix docker mount

chung1905:
Hello,

Error: `IsADirectoryError: [Errno 21] Is a directory: '.current_coin'`
Reason: If `.current_coin` doesn't exist, docker create a directory -> Read error.
Solution: move `.current_coin` and `.current_coin_table` to `data/current_coin` and `data/current_coin_table`.

With this, data will be stored in one place (easier to run on other machine), we have less hidden files.

Please check,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 *.log
-.current_coin
-.current_coin_table
 *.pyc
 __pycache__
 nohup.out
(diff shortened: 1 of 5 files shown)


chocobn69:
You might do something for users already using the docker system to upgrade from currrent files to new one

chung1905:
I think add the command to document is enough
```
mv .current_coin ./data; mv .current_coin_table ./data
```

chocobn69:
Sure

j-waters (approved):

PR #91, opened 2021-02-19
Move scout parameters into the config

sander0105:
I found myself playing around with the parameters for the Scouting routine so I moved them into the config to maintain the settings over time. 

- Moved the scout transaction fee and multiplier into the config and parse them
- Added new config to example config
- Added original scout configuration as defaults in case config keys are missing.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -27,6 +27,12 @@ USER_CFG_SECTION = 'binance_user_config'
 
 # Init config
 config = configparser.ConfigParser()
+config['DEFAULT'] = {
+    'scout_transaction_fee': '0.001',
+    'scout_multiplier': '5',
+    'scout_sleep_time': '5'
+}
+
 if not os.path.exists(CFG_FL_NAME):
     print('No configuration file (user.cfg) found! See README.')
     exit()
@@ -56,6 +62,10 @@ BRIDGE = Coin(BRIDGE_SYMBOL, False)
 # Prune settings
 SCOUT_HISTORY_PRUNE_TIME = float(config.get(USER_CFG_SECTION, 'hourToKeepScoutHistory', fallback="1"))
 
+# Get config for scout
+SCOUT_TRANSACTION_FEE = float(config.get(USER_CFG_SECTION, 'scout_transaction_fee'))
+SCOUT_MULTIPLIER = float(config.get(USER_CFG_SECTION, 'scout_multiplier'))
+SCOUT_SLEEP_TIME = int(config.get(USER_CFG_SECTION, 'scout_sleep_time'))
 
 class RequestsHandler(Handler):
     def emit(self, record):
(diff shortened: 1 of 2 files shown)


ghost:
What does the multiplier do and how would changing it impact how the bot runs?

sander0105:
The multiplier is used in conjunction with the transaction_fee setting to determine when to jump from one coin to the other. 

`        if (coin_opt_coin_ratio - transaction_fee * multiplier * coin_opt_coin_ratio) > g_state.coin_table[g_state.current_coin][optional_coin]:
`

CTetford:
Wouldn't a better method be to replace the multiplier with a profit parameter?

It's functionally the same, but a profit percentage like a "0.5% profit" would be an easier concept for people to work with rather than "multiplier of 5 x 0.1% transaction fee"

This way if someone is using BNB to pay for transactions (25% fee discount), they can configure it to 0.075% but still have a precise 0.5% profit without having an obscure multiplier of 6.667.

e.g.:
`if ((coin_opt_coin_ratio - transaction_fee * coin_opt_coin_ratio) / g_state.coin_table[g_state.current_coin][optional_coin]) > minProfit:`

Where minProfit = 1.01 for 1% profit

sander0105:
I agree, it makes sense to explore that direction but it would imply a fundamental change to the behavior of the script where my change was only meant to make changing the 2 scout parameters more accessible. I've been looking into configuring the profit parameter as well but that needs a bit more work and validation

edeng23:
I completely agree with @CTetford, I'll make this change in the near future. As of moving the parameters to the config file, I think that this is a good idea. @sander0105, can you rebase your code? 
Moreover, I think that the sleep time between scouts should also be moved out.

sander0105:
@edeng23 super, rebased my code and moved the scout timer as well

crypto_trading.py
```diff
@@ -26,10 +26,17 @@
 
 # Init config
 config = configparser.ConfigParser()
+config['DEFAULT'] = {
+    'scout_transaction_fee': '0.001',
+    'scout_multiplier': '5'
+}
```
rel-s, line 34: Perhaps we should add a default value for scout time as well:
`'scout_sleep_time': 5`

sander0105: Agreed, added :)

rel-s (approved):

rel-s:
Hi @sander0105, sorry for the wait! We wanted to merge some other stuff first. Could you rebase again so we can merge this PR?

crypto_trading.py
```diff
@@ -27,11 +27,22 @@
...
 if not os.path.exists(CFG_FL_NAME):
     print('No configuration file (user.cfg) found! See README.')
     exit()
 config.read(CFG_FL_NAME)
 
+# Get config for scout
+SCOUT_TRANSACTION_FEE = float(config.get(USER_CFG_SECTION, 'scout_transaction_fee'))
+SCOUT_MULTIPLIER = float(config.get(USER_CFG_SECTION, 'scout_multiplier'))
+SCOUT_SLEEP_TIME = int(config.get(USER_CFG_SECTION, 'scout_sleep_time'))
+
```
rel-s: I don't think these values are actually used in the code... Perhaps when you rebased the usage got lost?

j-waters:
Also 20 commits seems like an awful lot, might be best to squash them!

crypto_trading.py
sander0105: hmm must have messed up merging the last set of changes... let me check

sander0105: Working again :)

sander0105:
@j-waters cleaned up

nanashj:
#124

sander0105:
@nanashj can you elaborate on linking the #124 issue to this PR? I fail to see the connection.

j-waters (approved):

edeng23 (approved):

Issue #81, opened 2021-02-18
Web GUI / Machine Readable Log


j-waters:
[In my fork](https://github.com/j-waters/binance-trade-bot) I've added some machine readable logs to keep track of when trades happen, the value of the current coin in BTC and USD, and how close we are to jumping to another coin.

These logs can then be used to make graphs of gains/losses or monitor the real time performance of the bot. If this project continues and proves useful, I'd probably end up making a basic web interface using this data.

I was thinking of making a pull request but I thought I'd open an issue first to ask a couple of questions:

* What kind of data do we want to keep track of?
  * Current coin value in USD and BTC
  * Amount of this coin that you're holding
  * Current ratio with other coins and the ratio targets from the coin table
  * BTC value or general crypto market value? To determine the performance of the bot vs the entire market?
  * ... ?
* How do we want to store the data?
  * CSV - that's what I'm using at the moment
  * SQLight? As the project develops this may be better than having several text files keeping track of the state and stats

I'd be happy to write whatever and make a pull request, or leave it to you if there's a specific way you want to do things

rel-s:
Hi @j-waters! I think it's a great idea to have a web interface with some pretty data, and better logs. Using SQLite sounds like a good direction. Why don't you submit a PR when you have something ready?
We can discuss it in more detail in the Discord channel, but I think the data points you mentioned would be a great start. Just keep in mind we also have a Telegram bot "frontend", so we might want to integrate those somehow...
What do you think, @edeng23 ?

edeng23:
I definitely agree, I always wanted to have a "frontend" graph view so that I can tell if I'm getting any closer to trades instead of being in the dark. 

The thing about emphasising the current coin value in USD and BTC is that it confuses people and makes them think that the bot is supposed to constantly increase this parameter. This is not the point of the bot. Ideally, the GUI should show the amount of BTC or USD you would have if you didn't run the bot (if you were just holding the current coin the whole time), compared to the amount you have now.

diedline:
Have you considered testing whether historical data can be profitable。

ch3ckmat3:
The console can also be improved using Text UI for example: 
[https://www.willmcgugan.com/blog/tech/post/building-rich-terminal-dashboards/](https://www.willmcgugan.com/blog/tech/post/building-rich-terminal-dashboards/)

tkarakasoglu:
For now I have added a method, maybe you want to add this simple function. This method prints near 5 coins to current coin.

[...]

PR #93, opened 2021-02-19
Add database support

j-waters:
I was going to start work on #81 but decided to get some basic database support working first. All state is now held in an SQLite database. I've also made it a lot easier to add and remove coins on the fly. 

I've tested this on several full trades and it's working fine, but it could probably do with some more thorough testing. Also, I updated some of the config reading code to work with the new system but didn't look too hard at how it worked, so that could probably be simplified now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,9 @@ For more information about Telegram bots refer to [Telegram's official documenta
 
 ### Docker
 
-`docker-compose up`
+```shell
+docker-compose up
+```
 
 # Join the Chat
 
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ nohup.out
 user.cfg
 .idea/
 .vscode/
+venv/
+crypto_trading.db
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -421,7 +421,9 @@ def migrate_old_state():
             with db_session() as session:
                 for from_coin, to_coin_dict in table.items():
                     for to_coin, ratio in to_coin_dict.items():
-                        pair = get_pair(from_coin, to_coin)
+                        if from_coin == to_coin:
+                            continue
+                        pair = session.merge(get_pair(from_coin, to_coin))
                         pair.ratio = ratio
                         session.add(pair)
 
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -1,19 +1,19 @@
 #!python3
-from binance.client import Client
-from binance.exceptions import BinanceAPIException
-import logging
+import configparser
+import datetime
+import json
 import logging.handlers
 import math
-import time
 import os
-import json
-import configparser
-from logging import Handler, Formatter
-import datetime
-import requests
-import random
 import queue
+import random
+import time
+import traceback
+from logging import Handler, Formatter
 
+import requests
+from binance.client import Client
+from binance.exceptions import BinanceAPIException
 from sqlalchemy.orm import Session
 
 from database import set_coins, set_current_coin, get_current_coin, get_pairs_from, \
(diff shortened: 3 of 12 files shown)


j-waters:
The schema is very simple at the moment:

* `coins`
  * `symbol` - str, private key
  * `enabled` - bool
* `current_coin_history` - the most recent entry is the current coin
  *  `coin_id` - foreign key linking to coins table
  * `datetime`
* `pairs` - all the possible pairs
  * `from_coin_id` - foreign key, the 'current' coin
  * `to_coin_id` - the coin that you could move to from the other coin
  * `ratio` - required price ratio for the trade to occur

j-waters:
This basically adds no functionality, it's a straight port from the backup files to the database. I'd like to add a `trades` table, which would act as both a history of trades made as well as provide some crash resistance - we could keep track of the state of trades so that if the script is stopped before a trade is completed we can carry on.

And as I said in #81, we can have tables for keeping track of the current ratios between all the coins, current coin values, current market values etc that could then be used by whatever UI ends up being made

edeng23:
Looks good, I'll start testing it right now :)

porter707:
This is great, I've given this a little test (not made any trades with it) and the initial startup all appears to work and the Database looks good.

One request/suggestion would be an upgrade path for existing installs, so it detects an existing current_coin_table and loads that data into the pairs table. Then rename the current_coin_table file so it doesn't get picked up again on the next launch, then the user can manually remove that file if they are happy the upgrade went smoothly.

rel-s:
I agree with @porter707, it would save us a hefty amount of new issues to deal with...

edeng23:
@porter707 Good point, we wouldn't want existing users to lose their coin ratios.

j-waters:
Good idea I'll do that now

slyths:
@j-waters a bit off topic, but are you on the discord server and if you are, could you please tell me the nick you are using on discord. I would like to update your role.

rel-s:
Also worth noting that this PR makes #77 obsolete, so maybe we don't have to merge it.

j-waters:
Okay that should be working, I've kept the docker volume mounts so that migrating works for docker users although they will be able to be removed at some point.

As for #77 it might still be good to have one single config file instead of two

@slyths I am although I haven't looked at it very much! I'm Inazuma#4514

j-waters:
I think this also makes https://github.com/edeng23/binance-trade-bot/pull/85 obselete

chung1905:
Yes, going to close it now.
Since this project is getting big, I think it's important to do a proper code style.

README.md
```diff
@@ -56,7 +56,10 @@ For more information about Telegram bots refer to [Telegram's official documenta
 
 ### Docker
 
-`docker-compose up`
+```shell
+mkdir data
+docker-compose up
+```
```
chung1905: We can just create a .gitignore in data directory.
Can save some issue in repo.

j-waters: Great point

j-waters:
Been using this branch for a while now and haven't encountered any issues, I'd be happy for it to be merged. I've continued on from this in PRs #106 and #107

edeng23:
I saw, I'm probably going to merge it very soon. I'm nearly done with my tests.

edeng23:
@j-waters I found two bugs in the "migrate old state" logic. 
1. When you call session.add(pair) on line 426 of crypto_trading.py, I believe that you are trying to add a pair that is already in the database causing the following error:
_"sqlalchemy.exc.InvalidRequestError: Can't attach instance <Coin at 0x7fffeb9babe0>; another instance with key (<class 'models.coin.Coin'>, ('XLM',), None) is already present in this session."_
2.  The code crashes when trying to add coins that are referencing themselves in the old list (Coin A -> Coin A, ratio=1). I take the blame for writing ugly code in the first place. That being said, in order to make your code backwards compatible, you should bypass this when loading the old coin table.
Other than these two issues, everything else ran perfectly. Please fix these problems, I'm eagerly waiting to merge your code :)

j-waters:
Ah thanks, that should be fixed!

PR #94, opened 2021-02-19
Heartbeat log message

CTetford:
Add a heartbeat message every hour to indicate script is still functioning normally.


rel-s:
This should be, if existent at all, configurable. Wouldn't you rather have the ability to actively check if the bot is running somehow instead of receiving lots of messages?

CTetford:
Fair point, I just threw it together quickly because mine was sitting for hours at a time without doing anything. It can definitely be improved (and I agree, configurable). 

I've seen a few messages on Discord of people asking if it's doing anything, so this would be a stopgap for now until a better system is in place (like a frontend monitoring interface).

CTetford:
Updated to be configurable now and turned off by default.

thiazzz:
Testing this. Seems to work. Setting 3600.

rel-s:
We are currently working on a new API + web GUI, so this won't be necessary soon. For now, we won't merge this into master.

PR #95, opened 2021-02-19
Reformat with Black

j-waters:
[Black](https://github.com/psf/black) is a strict Python codestyle checker. All pull requests will be automatically checked with it to maintain a consistent code style.

Using Black is as simple as `pip install black` and then `black crypto_trading.py`. Most editors have an integration.

Fair enough if we don't want to add this, I was just running into issues where my IDE was trying to reformat stuff and it was causing a lot of extra unrelated changes to be committed, and since there are a lot of contributors it might be nice to have some sort of consistency. 

Also we could wait to commit this until after some recent pull requests have been merged, although rebasing should be simple enough

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,13 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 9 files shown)


j-waters:
Rebased

j-waters:
Closed in favour of #183

PR #101, opened 2021-02-20
Fix retry issue

maltoze:
Currently, `howmany` will never take effect when it is greater than `1`.

This will fix that.


rel-s:
Hi @maltoze, thanks for your help! The bug you found seems to be part of a bigger design issue, as specified in #115. We will soon open a new PR that tackles the entire issue. Thanks for pointing this out!

PR #106, opened 2021-02-21
Logging

j-waters:
This depends on #93 

Adds database tables to log trading, scouting and value histories.

I'm definitely open to changing what we log based on suggestions, or adding more tables. 

* Trade history
  * Logs each individual trade, so a trade from VIM to ETH would make two log entries, VIM->USDT and USDT->ETH
  * Stores the starting value of the two coins, and the amount traded, so we can easily infer the end value
* Scouting history
  * Logs for each pair we check during scouting, so that you can keep track of which coin it's getting close to switching to
  * Stores the pair, the ratio we're aiming for, the current coin price and the other coin price. Current ratio can then be inferred from the two prices
  * This creates `count(coins) - 1` log entries every 5 seconds, which if left could obviously start to use a fair amount of storage, so a method of pruning old entries will be a must. I haven't implemented this yet
* Value history
  * Logs the value of each enabled coin
  * Stores the coin, your current balance (if balance is zero, an entry is not made), the price of the coin in btc and the price of the coin in usd
    * I know the purpose of the bot isn't to maximise value but I think keeping track of this will be valuable to a lot of people
  * This will also need to be pruned, but probably not as aggressively as scouting history. We still want to preserve this history even if we don't have it at 5 second granularity. Maybe we don't even need to be logging it every 5 seconds. Maybe we have a second table for daily value logs. Maybe when we prune this table we leave one log entry for each coin per day. I don't know!

### DIFF
--- a/database.py
+++ b/database.py
@@ -2,7 +2,7 @@ from contextlib import contextmanager
 from datetime import datetime, timedelta
 from typing import List, Union, Optional
 
-from sqlalchemy import create_engine, and_
+from sqlalchemy import create_engine, func
 from sqlalchemy.orm import sessionmaker, scoped_session, Session
 
 from models import *
@@ -51,8 +51,7 @@ def set_coins(symbols: List[str]):
         for from_coin in coins:
             for to_coin in coins:
                 if from_coin != to_coin:
-                    pair = session.query(Pair).filter(
-                        and_(Pair.from_coin == from_coin, Pair.to_coin == to_coin)).first()
+                    pair = session.query(Pair).filter(Pair.from_coin == from_coin, Pair.to_coin == to_coin).first()
                     if pair is None:
                         session.add(Pair(from_coin, to_coin))
 
@@ -92,8 +91,7 @@ def get_pair(from_coin: Union[Coin, str], to_coin: Union[Coin, str]):
     to_coin = get_coin(to_coin)
     session: Session
     with db_session() as session:
-        pair: Pair = session.query(Pair).filter(
-            and_(Pair.from_coin == from_coin, Pair.to_coin == to_coin)).first()
+        pair: Pair = session.query(Pair).filter(Pair.from_coin == from_coin, Pair.to_coin == to_coin).first()
         session.expunge(pair)
         return pair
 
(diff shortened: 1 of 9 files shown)


AnasAlmomany (approved):
I like the modularity in the new structure.

porter707:
Some ideas on how you could maybe handle the pruning?

For Scouting history have something run every hour to clear out entries that are older than 1 hour. As it's probably not very useful information to know how close you were to a trade 8 hours ago, when the coin in the last few hours has just been falling.

For Value history maybe add a new column in the database "interval" (or a better name) that has the possible values monthly, weekly, daily, hourly

Monthly entries get deleted if they are older than a year
Weekly are deleted if they are older than a month
Daily are deleted if they are older than a week
Hourly get deleted if they are older than a day

models/coin_value.py
```diff
@@ -0,0 +1,49 @@
...
+    def usd_value(cls):
+        return cls.balance * cls.usd_price
+
+    @hybrid_property
+    def btc_value(self):
+        if self.btc_price is None:
+            return None
+        return self.balance * self.btc_price
+
+    @usd_value.expression
```
rel-s: Shouldn't this be:
```suggestion
    @btc_value.expression
```

database.py
```diff
@@ -105,6 +105,40 @@ def get_pairs_from(from_coin: Union[Coin, str]):
...
+def log_scout(pair: Pair, target_ratio: float, current_coin_price: float, other_coin_price: float):
+    session: Session
+    with db_session() as session:
+        pair = session.merge(pair)
+        sh = ScoutHistory(pair, target_ratio, current_coin_price, other_coin_price)
+        session.add(sh)
+
+
+class TradeLog:
+    def __init__(self, from_coin: Coin, to_coin: Coin, selling=True):
```
edeng23: Is there a special reason why selling is an optional parameter? Both selling and buying happen at the same frequency, and not passing the parameter explicitly can look confusing when reading the code, don't you think?

rel-s:
I agree with @porter707 about the value history, sounds like a good idea. About the scouting history, maybe we should just start with a new configuration option to determine how far back they are saved with some default value. Later on this can get more sophisticated.

models/coin_value.py
j-waters: Ah good catch

database.py
j-waters: Yep no reason at all I'll change it!

j-waters:
So I've implemented pruning pretty much in line with @porter707's suggestions, and added a config option as @rel-s suggested.

I also added the `schedule` library to simplify running methods at certain time intervals. This replaces the `while True` loop.

j-waters:
Rebased

PR #107, opened 2021-02-21
Api

j-waters:
Depends on #106 

Currently just have an endpoint for everything we log from #106.

Features that I should probably add:

 <s>Pagination</s> - I don't think it's really necessary at the moment
- [x] Filtering
- [x] Live updates (Server Sent Events? Websockets?)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,8 +18,8 @@ services:
     volumes:
       - ./data:/app/data
     ports:
-      - 5123:8000
-    command: gunicorn api_server:app -w 1 --threads 1 -b 0.0.0.0:8000
+      - 5123:5123
+    command: gunicorn api_server:app -k eventlet -w 1 --threads 1 -b 0.0.0.0:5123
 
   sqlitebrowser:
     image: ghcr.io/linuxserver/sqlitebrowser
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,8 @@ services:
     volumes:
     - ./user.cfg:/app/user.cfg
     - ./data:/app/data
+    ports:
+    - 8000:8000
 
   sqlitebrowser:
     image: ghcr.io/linuxserver/sqlitebrowser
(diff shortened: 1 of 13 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+python crypto_trading.py &
+gunicorn api_server:app -w 1 --threads 1 -b 0.0.0.0:8000
```
ryan-shaw: Might be better to add something like supervisorctrl/s6-overlay to deal with this, if crypto_trading.py crashes the container will continue as the gunicorn process is in the foreground. 

Better yet- run one service per container and add a second container to run the gunicorn

j-waters: Completely forgot we were using docker compose good shout I've added a separate container

AnasAlmomany (approved):
Great work man, Clean code is always appreciated

docker-compose.yml
```diff
@@ -3,8 +3,20 @@ version: "3"
 services:
   crypto-trading:
     build: .
+    image: binance_trader
     container_name: binance_trader
     volumes:
     - ./user.cfg:/app/user.cfg
```
aelbozie, line 8: I don't think this needs to be in volumes as the whole directory is copied

docker-compose.yml
```diff
@@ -3,8 +3,20 @@ version: "3"
...
-    - ./.current_coin:/app/.current_coin
-    - ./.current_coin_table:/app/.current_coin_table
+    - ./data:/app/data
```
aelbozie, line 10: Exactly this, makes sense to mount a whole directory not individual files

aelbozie (approved):

j-waters:
I haven't added pagination but I don't think it's necessary for now

j-waters:
Rebased

j-waters:
@edeng23 I'm sure we'll add to the API or change it in the future but would it be okay if it was merged soonish? Don't want to have to keep rebasing it is all!

edeng23:
@j-waters Sure,you're right. I'll start testing the code ASAP.

PR #109, opened 2021-02-21
Add dry run functionality

majvan:
At the beginning, the code checks for the file ".balance_coin_table"
where it stores the last dry-run balance of coins. If it does not exist,
it creates one from the current balance read from Binance.

Then it provides an update to the balance table after every sell/buy
is executed and pretends it was executed successfully.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -17,6 +17,7 @@ import queue
 # Config consts
 CFG_FL_NAME = 'user.cfg'
 USER_CFG_SECTION = 'binance_user_config'
+DRY_RUN = True
 
 # Init config
 config = configparser.ConfigParser()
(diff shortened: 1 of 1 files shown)


rel-s:
Hi @majvan, great idea! I think that in it's current form, this PR introduces too much clutter into the main project file which will in turn make it cumbersome to maintain... We should think of a better way to incorporate this idea into the project. Perhaps this can be a different file (`crypto_trading_sim.py`)? Do you have other ideas? @edeng23

edeng23:
I stand behind @rel-s on this, I think that the code becomes too cluttered with different concepts. Perhaps there is a more subtle way to incorporate this idea rather than to "if" it all in. 
Other than that, I think that as a new feature it is really great.

majvan:
Hi guys.
From a viewpoint of a clean / proper code, we should add a class "Trader" or "Strategy" to decouple the logic from the scan.
Then there should be another class for "Exchange" which would hold stateful data regarding exchange.
In the case of the dry run, a DryRunMixin attached to an Exchange class would make its job.

This 'redesign' is far-away from my code. I wanted to stick with narrow procedural approach like it is done in this project. If we want to decouple and correct logic, we have to move to classes and mixins / compositions.

Issue #115, opened 2021-02-22
Fix Potential Hang on Transaction


rel-s:
It seems like the `while` loops in `transaction_through_tether` are redundant, because both `sell_alt` and `buy_alt` are protected with `@retry`.
If the loops stay, we might attempt to buy altcoins when they are at a different value than the one intended, perhaps due to a failure on Binance's server side.
Required fixes:

- [x] Remove `while` loops in `transaction_through_tether`.
- [x] `@retry` should return a `None` value on failure, and this value should checked.
- [x] Make sure we are not logging these timed-out transactions as successful.

rel-s:
If we managed to sell but not to buy, we are in a bad state. We can't bail on a failed buy, since we will get stuck on the bridge coin. We need to implement better logic for escaping from a bridge coin, but this is out of scope for this issue.

rel-s:
Fix merged, closing for now

PR #116, opened 2021-02-22
Fix Retry Bug

rel-s:
`@retry` decorator fixed, and we implement a bail from failed sales. We don't bail from a failed buy, since this required more sophisticated logic.


rel-s:
Fixes issue #115.

Issue #118, opened 2021-02-23
Switch to Websocket API


rel-s:
To allow a higher api request volume, we should use Binance's websocket APIs instead of the current implementation.

luponata:
If i understand correctly (new repo user), we (should) or already use websockets to receive tickers, balance updates, orders updates... and API calls to make or cancel them.

ATM as far as i know, binance doesn't have a roadmap to allow making orders with websockets

https://dev.binance.vision/t/sending-and-canceling-orders-from-websocket-connection/437
https://www.binance.com/en/support/announcement/360035944992

Were there any developments unknown to me?

jshield:
yeah so you can't place them, but you can pretty much subscribe to streams for everything else.

PR #97, opened 2021-02-19
Apprise support

porter707:
This adds support to send notifications to multiple different services (telegram, discord, slack, email etc) via Apprise

PR #119, opened 2021-02-23
Binance API Manager

porter707:
This branch is aimed to be a start on https://github.com/edeng23/binance-trade-bot/issues/118 by refactoring the Binance code into it's own class to make future improvements easier.

It currently depends on the logging class from my other branch https://github.com/edeng23/binance-trade-bot/pull/97 which just makes using the same logger across different python files easier, if you don't like the way that logging class is implemented I'm happy to implement any suggestions.

I've been trying to test this locally by manually calling the various functions (but not actually making any buys/sells) and it all appears ok, I'm waiting on  an actual buy/sell to happen which may be a while with the current dip.

### DIFF
--- a/binance_api_manager.py
+++ b/binance_api_manager.py
@@ -1,0 +1,209 @@
+from binance.client import Client
+from binance.exceptions import BinanceAPIException
+from database import TradeLog
+from models import Coin
+from logger import Logger
+import math, requests, time
+
+
+class BinanceAPIManager:
+    def __init__(self, APIKey: str, APISecret: str, Tld: str, logger: Logger):
+        self.BinanceClient = Client(APIKey, APISecret, None, Tld)
+        self.logger = logger
+
+    def get_all_market_tickers(self):
+        """
+        Get ticker price of all coins
+        """
+        return self.BinanceClient.get_all_tickers()
+
+    def get_market_ticker_price(self, ticker_symbol: str):
+        """
+        Get ticker price of a specific coin
+        """
+        for ticker in self.BinanceClient.get_symbol_ticker():
+            if ticker[u"symbol"] == ticker_symbol:
+                return float(ticker[u"price"])
+        return None
+
+    def get_currency_balance(self, currency_symbol: str):
+        """
+        Get balance of a specific coin
+        """
+        for currency_balance in self.BinanceClient.get_account()[u"balances"]:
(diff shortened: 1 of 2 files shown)


binance_api_manager.py
```diff
@@ -0,0 +1,202 @@
...
+from crypto_logger import Logger
+from models import Coin
+import math, requests, time
+
+
+class BinanceAPIManager:
+
+    APIKey = None
+    APISecret = None
+    Tld = None
```
rel-s: I don't think these members are necessary, they aren't really used anywhere outside __init__()

binance_api_manager.py
```diff
@@ -0,0 +1,202 @@
...
+
+    BinanceClient = None
+
+    def __init__(self, APIKey, APISecret, Tld):
+
+        self.APIKey = APIKey
+        self.APISecret = APISecret
+        self.Tld = Tld
```
rel-s: Same goes here, we can just pass these values on to Client() without saving them

rel-s:
We would like to merge this PR soon, sorry for the wait! Could you rebase this? At the current state it's a bit hard to review as it's not entirely clear what changes are because of this PR and what are due to #97. Thanks!

porter707:
@rel-s this is rebased now and ready for review

edeng23:
@porter707 Thanks :) I'll run a few tests before merging.

binance_api_manager.py
```diff
@@ -0,0 +1,209 @@
...
+                )
+                time.sleep(1)
+            except BinanceAPIException as e:
+                self.logger.info(e)
+                time.sleep(2)
+            except Exception as e:
+                self.logger.info("Unexpected Error: {0}".format(e))
+
+        newbal = self.get_currency_balance(alt_symbol)
+        while newbal >= bal:
```
edeng23: I got this error after running your code:
2021-02-26 21:12:06,030 - crypto_trader_logger - INFO - Failed to Buy/Sell. Trying Again.
2021-02-26 21:12:06,032 - crypto_trader_logger - INFO - name 'bal' is not defined

I believe that "bal" should be replaced with "alt_balance" in accordance with @j-waters's revision.

porter707: Ahh yes, sorry thats fixed now

PR #122, opened 2021-02-23
Picking Best Pair Instead of First

rel-s:
This changes the scouting logic so that it selects the best pair to trade, instead of the first value.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -365,9 +365,9 @@ def scout(client, transaction_fee=0.001, multiplier=5):
     if ratio_dict:
       max_optional_coin = max(ratio_dict, key=ratio_dict.get)
       logger.info('Will be jumping from {0} to {1}'.format(
-            g_state.current_coin, optional_coin))
+            g_state.current_coin, max_optional_coin))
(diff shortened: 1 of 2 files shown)


crypto_trading.py
```diff
@@ -396,12 +398,19 @@ def scout(client: Client, transaction_fee=0.001, multiplier=5):
...
+    # keep only ratios bigger than zero
+    ratio_dict = dict(filter(lambda x: x[1] > 0, ratio_dict.items()))
+
+    # if we have any viable options, pick the one with the biggest ratio
+    if ratio_dict:
+      max_optional_coin = max(ratio_dict, key=ratio_dict.get)
+      logger.info('Will be jumping from {0} to {1}'.format(
+            current_coin, max_optional_coin))
+      transaction_through_tether(
+          client, current_coin, max_optional_coin)
```
maltoze, line 413: seems have a inconsistent indentation.

maltoze: seems have a inconsistent indentation.

siefgred (approved):
LGTM

PR #123, opened 2021-02-24
Refactor

j-waters:
This is a biggie. I don't expect it to be merged (although I'd like it to be) since it is rather nebulous, but essentially I've refactored the entire project.

This depends on #106, is inspired by #116, #119, and supersedes #95.

This splits the entire project into several classes. There are no more global variables. The entire project is then within its own module, which separates it from the config files, data, docker stuff, etc. The entire project has then been put through a formatter to make the code style consistent.

Splitting into classes allows for things like dry runs, backtesting, alternative logging methods, alternative exchanges, alternative strategies to be placed in a lot more easily. It also makes the code a hell of a lot cleaner.


.github/workflows/black.yml
```diff
@@ -0,0 +1,13 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
```
sander0105, line 7: I would recommend pinning it to 20.04 
also see: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

.user.cfg.example
```diff
@@ -6,3 +6,4 @@ bridge=USDT
 botChatID=
 botToken=
 tld=com
+hourToKeepScoutHistory=1
```
sander0105, line 9: The scheduler set up for this runs every minute. Wouldn’t it make sense to also configure this in minutes?

binance_trade_bot/binance_api_manager.py
```diff
@@ -0,0 +1,186 @@
...
+
+        while stat["status"] != "FILLED":
+            try:
+                stat = self.client.get_order(
+                    symbol=alt_symbol + crypto_symbol, orderId=order["orderId"]
+                )
+                time.sleep(1)
+            except BinanceAPIException as e:
+                self.logger.info(e)
+                time.sleep(2)
```
sander0105, line 77: There a multiple sleep(x)'s here, it feels arbitrary but I suspect otherwise. Is it worth pulling these into variables to make more sense of this?

sander0105:
You mentioned this being a biggie but looking through it, it seems a step forward worth taking. This will definitively allow for easier extension in the future!

.github/workflows/black.yml
j-waters: Yep can do!

.user.cfg.example
j-waters: Yeah that's fair, that part is originally from #106 so I can change it there

binance_trade_bot/binance_api_manager.py
j-waters: Honestly I just copied a pasted a lot of the transaction code without modifying it, the sleeps were there already. Someone will look into how necessary all of them are at some point I'm sure

sander0105: Fair point :)

j-waters:
I'll close this in favour of smaller PRs, the first of which is #129.

PR #129, opened 2021-02-25
[refactor] Make project a package

j-waters:
Now we've got more than one Python file I think it's time to put everything into a package, it also separates the code from the data and configs.


j-waters:
Superseded by #148

PR #134, opened 2021-02-26
Don't keep attempting transaction if the amount is invalid

### LINKED ISSUE #75: Failed to Buy/Sell. Trying Again.
2021-02-17 21:22:05,578 - crypto_trader_logger - INFO - Selling 0.0 of BTT
2021-02-17 21:22:05,692 - crypto_trader_logger - INFO - Balance is 0.0
Failed to Buy/Sell. Trying Again.

I have only 0.00118869 BTC in my acc. But it's giving this issue :/

j-waters:
I feel like eventually we should remove the big retries and handle as many types of errors individually

Maybe we should have a check in the scouting code to see if we actually have enough of the current coin. I was thinking of adding a method that would check through all the coins you have enabled and switch to the one with the biggest bridge coin value in the scenario where you don't have enough of the current coin. This could also be how we set the initial coin if the user doesn't specify it, instead of selecting one randomly. 

Kind of fixes #75

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -238,6 +238,9 @@ def buy_alt(client: Client, alt: Coin, crypto: Coin):
             order_recorded = True
         except BinanceAPIException as e:
             logger.info(e)
+            if e.code == -1013:
+                logger.info(f"Can't execute transaction: {order_quantity} is an invalid amount")
+                return None
             time.sleep(10)
         except Exception as e:
             logger.info("Unexpected Error: {0}".format(e))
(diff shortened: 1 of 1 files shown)


j-waters:
Kind of fixes #75

crypto_trading.py
```diff
@@ -238,6 +238,9 @@ def buy_alt(client: Client, alt: Coin, crypto: Coin):
             order_recorded = True
         except BinanceAPIException as e:
             logger.info(e)
+            if e.code == -1013:
+                logger.info(f"Can't execute transaction: {order_quantity} is an invalid amount")
+                return None
```
rel-s, line 243: Currently, returning a None value from buy_alt() doesn't really quit the buy state, because the call is in a while loop.

j-waters:
I'll close this in favour of #137

PR #136, opened 2021-02-26
✨ General Optims

segersniels:
### Added
- `.editorconfig` so people have a basic form of linting as it's supported natively by most code editors
- Basic environment variable support through `entrypoint.sh` substituting config values in `user.cfg`
- Initial Github Actions setup (requires `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets to be configured) to build and push the image to DockerHub (multiarch). This way users can more easily get started without having to pull the entire repo each time to build from scratch.

### Changed
- Ran basic linting on the main script for some better consistency
- Cleaned up `Dockerfile` and used `slim` tag to reduce size

### Example `docker-compose.yml`
```yml
version: "3"

services:
  binance:
    build: .
    environment:
      SUPPORTED_COINS: "BTC ETH ADA BNB LTC XMR XLM ATOM EOS TRX IOTA VET NEO DASH ETC NANO ICX SUSHI ALGO UNI CAKE"
      BRIDGE_CURRENCY: USDT
      CURRENT_COIN:
      API_KEY:
      API_SECRET_KEY:
    volumes:
      - binance-data:/app
    restart: always

volumes:
  binance-data:
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# top-most EditorConfig file
+root = true
+
+# Default ruleset for all files
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,6 +21,7 @@ jobs:
         id: docker_build
         uses: docker/build-push-action@v2
         with:
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ github.repository }}:latest
       - name: Image digest
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,27 @@
+name: binance-trade-bot
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: ${{ github.repository }}:latest
+      - name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,20 @@
-FROM python:3.8 as base
-
-FROM base as builder
-
-RUN mkdir /install
+FROM python:3.8 as builder
 
 WORKDIR /install
 
 COPY requirements.txt /requirements.txt
 
-RUN pip install --prefix=/install -r /requirements.txt
+RUN pip install --no-cache-dir --prefix=/install -r /requirements.txt
 
-FROM base
+FROM python:3.8-slim
 
 WORKDIR /app
 
-COPY --from=builder /install /usr/local\
-COPY . /app
+COPY --from=builder /install /usr/local
+
+COPY . .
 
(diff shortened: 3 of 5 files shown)


segersniels:
If enough people are interested in this change I suppose I can also add ENV vars for Apprise.

j-waters:
This is very much needed! I like everything that's here and would be happy for it all to be merged however it might be best to make separate pull requests for each feature, @edeng23 might be happy with it as is though

segersniels:
Also, to repeat my discord question in #bugs-and-support to whomever it may concern:

> Wouldn't it be better to `raise` the scouting error instead of halting process termination by putting it in a `try except`? That way docker can restart the container if it's experiencing issues.

Been running the bot for little over a week now and at a certain point the websocket connection was lost and it kept logging scouting errors without actually exiting the process. Only way currently to fix this is to manually restart the bot. Would find it more helpful to actually retry to get a valid connection or just `raise` the error so docker can handle restarting.

BelphegorPrime:
There are some users that run it also on older Raspberry Pi´s as I read the lowest should be some Raspberry Pi 3 and Raspberry Pi Zero W. Fast googling shows that they are ARMv6 and ARMv7.
Can you add support for them too?

bastdus:
what about people who's not using docker ?

segersniels:
> what about people who's not using docker ?

This PR is specifically for docker. However it could be possible to refactor the actual script to prefer ENV vars over config values. 

On the other hand I don't really see a benefit for people that are trying to run it without docker and want to manually setup anyway? I could be wrong here however... 

I have 0 Python experience though so might be better if somebody else has a shot at it to refactor the script.

BelphegorPrime:
From my experience in the discord, I can say that there isn't a negligible amount of users running it natively.

bastdus:
I'm not using docker (I mean not on this way for this project). Here I prefer start an detach unbuntu container. And I go inside interactively with bash. Like that my container are always run, and no need to kill and rebuild 2 containers at every update. I can add some stuff without any problem : for exemple => cron that send me resume of trade at 20:00, runing 2 or 3 bot if I want, test different commit, got my own git repo, modify code...etc...etc...I think you got it ;)

jeremiesigrist:
> I'm not using docker (I mean not on this way for this project). Here I prefer start an detach unbuntu container. And I go inside interactively with bash. Like that my container are always run, and no need to kill and rebuild 2 containers at every update. I can add some stuff without any problem : for exemple => cron that send me resume of trade at 20:00, runing 2 or 3 bot if I want, test different commit, got my own git repo, modify code...etc...etc...I think you got it ;)

I confirm, I use it the same way

segersniels:
> I'm not using docker (I mean not on this way for this project). Here I prefer start an detach unbuntu container. And I go inside interactively with bash. Like that my container are always run, and no need to kill and rebuild 2 containers at every update. I can add some stuff without any problem : for exemple => cron that send me resume of trade at 20:00, runing 2 or 3 bot if I want, test different commit, got my own git repo, modify code...etc...etc...I think you got it ;)

At that point why are you even using docker haha. Just run your scripts in different `screen` instances and use the systems crontab to gather whatever you need? But to each their own 🤷‍♂️ 

But I do agree that an all around cleaner approach to supporting environment variables would be to adjust the python script instead of substituting config files on container startup.

bastdus:
> > I'm not using docker (I mean not on this way for this project). Here I prefer start an detach unbuntu container. And I go inside interactively with bash. Like that my container are always run, and no need to kill and rebuild 2 containers at every update. I can add some stuff without any problem : for exemple => cron that send me resume of trade at 20:00, runing 2 or 3 bot if I want, test different commit, got my own git repo, modify code...etc...etc...I think you got it ;)
> 
> At that point why are you even using docker haha. Just run your scripts in different `screen` instances and use the systems crontab to gather whatever you need? But to each their own 🤷‍♂️ 
> 
> But I do agree that an all around cleaner approach to supporting environment variables would be to adjust the python script instead of substituting config files on container startup.

I'm web dev, so my server is use 90% for that. I don't want mix my pro and my personal use. But it's not the topic here ;)

segersniels:
Closing this PR as I'm going to refactor the ENV var support by adjusting the python script itself and splitting up in multiple PR's per functionality.

https://github.com/edeng23/binance-trade-bot/pull/142 and https://github.com/edeng23/binance-trade-bot/pull/143

PR #130, opened 2021-02-25
Remove duplicated code

j-waters:
Move duplicated code into separate methods

PR #137, opened 2021-02-26
Allow multiple coins to be active at once, recover from failed buy

### LINKED ISSUE #75: Failed to Buy/Sell. Trying Again.

### LINKED ISSUE #135: Idea: Remove concept of 'current coin'
I was thinking about this after making #134.

Is keeping track of the 'current coin' even necessary? The scout method could be changed to work as follows:

1. Filter the enabled coins to find ones that have a high enough balance to be sold
2. For each of those coins, use the current scouting algorithm

I think that would be simple enough to do, and would fix issues like #75 better. 

This would then allow for the risk to be spread. At the moment as per #122 if there are multiple possible coins we could switch to, we pick the best one. If there was no concept of the 'current coin', we could instead switch from one coin to multiple different coins, and spread the risk. Or when switching to a coin, we could just convert half of the current coin, incase the current coin continues to spike.

I'd happily work on this if people think it's a good idea

j-waters:
Removes the concept of the 'current coin' and instead searches for all enabled coins that have a balance large enough to be traded. More info in #135

Depends on #130 

Supersedes #134 

Closes #135 and closes #75


quantumsheep:
Having a `current_coins` file for initialization would be great along with this modification, what do you think?

j-waters:
Rebased

mehdico:
Not buying anything for me.

j-waters:
@mehdico it's been a while since I made this so it might be broken I'll have to retest it

@edeng23 is this something you'd consider merging? If so I'll rebase and add a method to buy a random coin if you only have a bridge coin - this would also allow us to recover from failed transactions a lot better

leon1995:
Currently, the bot is unusable for me, because I get stuck in a loop, where the bot is trying to sell/buy a coin but the api response with a 1013 error. As mentioned in this pr, this would be fixed here. Therefore, I would be very happy when this pr gets merged soon.

Thanks for your amazing work btw! 😃

jackMort:
@edeng23 any update here?

j-waters:
Rebased

Also added a method to automatically buy a random coin if all we have is the bridge coin, this allows us to better recover from failed transactions

Eventually this could be made smarter by picking a coin that we wouldn't immediately trade out of, by looking at the current ratios and current prices, but I haven't really thought about how that logic would work

johannes-peeters:
> Rebased
> 
> Also added a method to automatically buy a random coin if all we have is the bridge coin, this allows us to better recover from failed transactions
> 
> Eventually this could be made smarter by picking a coin that we wouldn't immediately trade out of, by looking at the current ratios and current prices, but I haven't really thought about how that logic would work

I implemented this logic in my #199 PR. I'm no longer keeping from/to ratio, but compare latest trades to latest prices

j-waters:
Added a method to scout for the optimal coin to buy with the bridge coin - the bot will find the one coin that has all negative ratios and buy that. Buying any other coin would cause the bot to immediately trade away from it, wasting fees.

edeng23:
@j-waters since the code has now been modularized, do you think that we can implement this concept as an alternative to auto_trader.py?
The way I see it, in the future we may have a few different trading strategies and having a different "main" python file for each one can be clean and easy to use. Perhaps we should let the user choose between strategies on startup.

j-waters:
@edeng23 that's an excellent point, I'll do that

johannes-peeters (commented):
Shouldn't we make this a config option? People might start the bot without reading proper documentation, and end up with a coin when they're not really committed yet.

j-waters:
So I'll leave recovering from a failed buy in the default strategy I think. I would personally say that this entire PR could be part of the default strategy, since it doesn't actually change the behaviour too much - sure it allows for multiple coins to be active at one, but as it stands only one ever will be. I'll still open another PR though if you'd prefer it to be a separate strategy @edeng23, I've already written some code for configuring a strategy to use

j-waters:
Done. I still think this behaviour should be the default, but I've opened https://github.com/edeng23/binance-trade-bot/pull/204 which adds strategy selection as well

leon1995:
I still get the error that it failed to buy/sell
```
Selling 0.0 of ICX
Balance is 0.00404
Failed to Buy/Sell. Trying Again.
```

How can I fix this?

j-waters:
@leon1995 that really shouldn't be possible, so I'm not sure, sorry.

leon1995:
@j-waters I think I found the issue. The bot traded my current_coin and after a restart, the bot could not trade it, because my balance was to low.
When I change the current_coin to the main coin I have in binance, before starting the bot, it works fine. But this means, I have to change it on every restart... is there a better solution?

j-waters:
@leon1995 are you sure you're using this branch? Because this branch completely removes the concept of 'current coin', and will only trade with coins where your balance is high enough - the issue you're describing is the exact issue this PR is design to fix

Also, it should be noted that this specific PR is unlikely to be merged, try #204

leon1995:
@j-waters I am using the master branch. I will switch to [this](https://github.com/j-waters/binance-trade-bot/tree/remove_current_coin) branch and try to work it out! thx

EDIT: How can I get rid of these small leftovers of the bot? On some coins I have a balance of just a very few coins with 0.02€ or so...

leon1995:
@j-waters I have no problems since I use your branch. However I discovered, that the bot tries to sell 0.0 coins over and over again
```
Selling 0.0 of VET
Balance is 0.76
Selling 0.0 of VET
Balance is 0.76
Selling 0.0 of VET
Balance is 0.76
```
Do you have a clue why this happens?

PR #138, opened 2021-02-26
Empty lines and trailing spaces removal 

PrzemyslawSarnacki:
Fix for recently observed bug which causes 'USDTUSDT' pair. Trailing spaces are stripped and every empty string is removed from the list of coins.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -54,7 +54,8 @@ supported_coin_list = []
 
 # Get supported coin list from supported_coin_list file
 with open('supported_coin_list') as f:
-    supported_coin_list = f.read().upper().splitlines()
+    supported_coin_list = f.read().upper().strip().splitlines()
+    supported_coin_list = list(filter(None, supported_coin_list))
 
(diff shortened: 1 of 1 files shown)


edeng23:
Hey, can you please rebase the code instead of using a merge commit? Thanks :)

PrzemyslawSarnacki:
Sure everything should be fine now. Sorry for that

j-waters:
This doesn't appear to do anything? You may have lost some code in the rebase

PrzemyslawSarnacki:
> This doesn't appear to do anything? You may have lost some code in the rebase

What do you mean by this? It does exactly what I've described. It is a fix for trailing lines inside coin list file (which results in a "" string inside coin list which in turn makes 'USDTUSDT' pair. Only two lines are changed to accomplish that.

j-waters:
Oh sorry I completely missed the `.strip()`! That is a good change although it still shouldn't fix the `USDTUSDT` pair issue, I opened a new PR that should fix that #139

PR #148, opened 2021-02-27
Modularise code

j-waters:
Added the Database class, AutoTrader class, and Config class, as well
as a utils file.

The benefit of doing this is twofold:

1. Code is clearer - split over multiple files, less reliance on global variables, less passing of the same objects all over the place
2. Allows for different parts of the app to be swapped out easily

Number 2 is the reason why I really want this merged. I really want to add some dry run or backtesting functionality like #109, and I think it would be a lot easier if I could just swap out the database for an in memory database, the Binance api manager for a fake one that doesn't actually make any trades, and the config object for one with different values.

On top of that we could then allow people to make their own strategies without having to modify everything - you could just extend AutoTrader and use that instead, making pulling changes from master a lot easier, and sharing new strategies a lot easier as well

I've also made the project into a package


j-waters:
I still think this would be really good to do, I'll wait until #154 is merged before I rebase though

rel-s:
Looks great to me! #154 is merged, can you rebase please?

j-waters:
@rel-s rebased

edeng23:
I like it too, Looks great!
I'll start running some tests and merge it soon enough :)

binance_trade_bot/crypto_trading.py
```diff
@@ -0,0 +1,38 @@
...
+    config = Config()
+    db = Database(logger, config)
+    manager = BinanceAPIManager(config, db, logger)
+    trader = AutoTrader(manager, db, logger, config)
+
+    logger.info("Creating database schema if it doesn't already exist")
+    db.create_database()
+
+    db.set_coins(config.SUPPORTED_COIN_LIST)
+    db.migrate_old_state()
```
edeng23, line 25: When do you think that this should be removed?

edeng23 (approved):

binance_trade_bot/crypto_trading.py
j-waters: It could probably be removed now tbh! Although there's no harm in keeping it, maybe if we do something that would require the migration code to be changed we just remove it then

PR #149, opened 2021-02-27
add message, that bot started scouting

BelphegorPrime:


edeng23:
I think that this PR will be obsolete after we merge the heartbeat PR (#184). Closing.

PR #151, opened 2021-02-28
🐳 Docker: Cleanup

segersniels:
### Fixed
- Don't remove default `CMD` from Dockerfile to prevent breaking older versions of trade bot. You can override it if you want by passing a different command...

### Changed
- Using `python:3.8-slim` as base image to reduce image size
- Cleaned up general Dockerfile for better readability

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,4 @@
-FROM python:3.8 as base
-
-FROM base as builder
-
-RUN mkdir /install
+FROM python:3.8 as builder
 
 WORKDIR /install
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,19 +1,16 @@
-FROM python:3.8 as base
-
-FROM base as builder
-
-RUN mkdir /install
```
segersniels, line 5: `WORKDIR` already creates the directory, no need to mkdir before you set it

Dockerfile
```diff
@@ -1,19 +1,16 @@
...
+FROM python:3.8-slim
 
-COPY --from=builder /install /usr/local
+WORKDIR /app
 
-COPY . /app
+COPY --from=builder /install /usr/local
+COPY . .
 
-WORKDIR /app
```
segersniels, line 19: Always add a basic command. It can be overridden anyway but at the very least doesn't require people to figure out which command to add to their docker container

PR #154, opened 2021-03-01
Code Cleanup

rel-s:
Made some changes so that code will be more consistent - most substantially moving some logic out of `main()` and changing variable names in the buy-sell functions to something less cryptic.

### DIFF
--- a/crypto_trading.py
+++ b/crypto_trading.py
@@ -162,7 +162,6 @@ def scout(client: BinanceAPIManager, transaction_fee=0.001, multiplier=5):
     '''
     Scout for potential jumps from the current coin to another coin
     '''
-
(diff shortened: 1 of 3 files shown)


rel-s:
Tested on my end successfully.

crypto_trading.py
```diff
@@ -133,11 +135,34 @@ def initialize_trade_thresholds(client: BinanceAPIManager):
             pair.ratio = from_coin_price / to_coin_price
 
 
+def initialize_current_coint(client: BinanceAPIManager):
```
porter707: is the "t" supposed to be on the end of this function ?

rel-s: It's not! Will fix soon

PR #157, opened 2021-03-01
config handling moved to config

Pekafu:
Not sure if these are actually worth or needed, but could be used to enable more than one instance running


j-waters:
I think it would be good to do something like this, I've done something similar in #148

PR #160, opened 2021-03-02
Display the spot balance after sold coin

llovvoll:
So that user can know the current spot balance in the message, and can clearly know whether it is losing money

image


rel-s:
The Bot's purpose is not to earn more USD, please see the README. We probably won't add this information at this point, as it can be misleading.

PR #161, opened 2021-03-02
Improvement on displaying on the terminal console activity

ardgraft:
Improvement to the console terminal activity to show the user the coin pairs being tested and the bridge coin. This is so users can see *some* activity and not think the bot has stopped/crashed/hung etc.


ardgraft:
A newer and better concept exists in PR #169 
I will close this one.

PR #162, opened 2021-03-02
Fix "Create a .ini file" in readme.md

Mystikal91:
The readme says that an .ini file should be created, while it should be a .cfg file, this is wrong and can cause some confusion


edeng23:
Agreed :)

PR #167, opened 2021-03-03
check max sell amount before ratio diff

sonerokur:
Checks "max sell amount greater than current buyable amount" to prevent loss.

### DIFF
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -107,19 +107,24 @@ class AutoTrader:
         '''
         Scout for potential jumps from the current coin to another coin
         '''
-        all_tickers = self.manager.get_all_market_tickers()
-
         current_coin = self.db.get_current_coin()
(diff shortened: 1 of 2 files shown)


j-waters:
Forgive me if I'm misinterpreting you PR, but this will make it so that the bot will never buy less of a coin than it previously held. So if yesterday it had sold 50 XMR, and today it wants to buy 40 XMR, it wouldn't.

What if you add or remove coins from your account? If I reduce the total value of my account it will never buy anything ever again because I won't have enough coins to beat my previous trades. If I increase the value of my account I'll always beat the value of my previous trades so it won't help.

The bot currently uses a ratio to avoid this problem, so if there is an issue with the bot losing coins surely we should fix that instead of doing this. Plus if we ever introduce the ability to have multiple current coins like in #137 this also wouldn't work.

Again I may have completely misinterpreted what your PR does so correct me if I'm wrong

sonerokur:
I didn't think about adding or removing coins. Completely agree with your thoughts. 

I'll try to investigate more to find more general solution.

PR #169, opened 2021-03-04
Configurable heartbeat message with variables

Brackyt:
This adds a heartbeat message with a configurable message that is sent every set amount of time (in seconds).
The message uses variables that can show different informations.
It includes the following variables:

- current_coin : Displays the current coin's symbol.
- balance: Displays the balance of the current coin.
- ratios: Displays the ratios in percentage for each pair with the current coin. With the default settings, the bot should switch coins when one of the pairs shows 0.5%. (hard to explain for me but see output below)

With this example heartbeat message: "Scouting. Current coin: {current_coin} - Balance: {balance}. Current ratios: \n{ratios}"
It will log this for example:

```
Scouting. Current coin: ETH - Balance: 0.0246. Current ratios: 
ETH to XLM. Diff: -0.23%
ETH to TRX. Diff: -0.63%
ETH to ICX. Diff: -0.79%
ETH to EOS. Diff: -0.54%
ETH to IOTA. Diff: -0.85%
ETH to ONT. Diff: -0.53%
```

If you have any other variable ideas that could be added I can add them.


Pekafu:
this seems nice, should be rebased and fixed to cope with the current version

VeriDigital:
awesome we need this.

Brackyt:
> this seems nice, should be rebased and fixed to cope with the current version

I will rebase it as soon as I can

crypto_trading.py
```diff
@@ -284,6 +338,10 @@ def main():
     schedule.every(1).minutes.do(update_values, client=client).tag("updating value history")
     schedule.every(1).minutes.do(prune_scout_history, hours=SCOUT_HISTORY_PRUNE_TIME).tag("pruning scout history")
     schedule.every(1).hours.do(prune_value_history).tag("pruning value history")
+    if HEARTBEAT_DURATION > 0:
+        schedule.every(HEARTBEAT_DURATION).seconds.do(
+            logger.info, HEARTBEAT_MESSAGE.format(**get_heartbeat_formatting(client=client))
```
sonerokur: get_heartbeat_formatting is called only at startup. after that, it always logs the same message.

Brackyt: Oops, I'll fix that. Thanks for noticing !

crypto_trading.py
```diff
@@ -257,6 +262,55 @@ def migrate_old_state():
...
+
+        current_value = (
+            coin_opt_coin_ratio
+            - SCOUT_TRANSACTION_FEE
+            * SCOUT_MULTIPLIER
+            * coin_opt_coin_ratio
+        )
+        difference = (current_value - pair.ratio) / pair.ratio * 100
+
+        heartbeat_msg += f"{current_coin.symbol} to {pair.to_coin.symbol}. Diff: {round(difference, 2)}%\n"
```
sonerokur: Can you add some aligning in here ? Probably, this will work.
`
f"{current_coin.symbol:<5} to {pair.to_coin.symbol:<5} Diff: {round(difference, 2):>6.2f}%\n"
`

Brackyt:
Rebased

Brackyt:
Messed up the rebase, redoing the PR from zero. Now here: #184

PR #174, opened 2021-03-04
🐳 Docker: Point to official image

segersniels:
### Changed
- `docker-compose.yml` example should use official image from now on


Pekafu:
probably should also update the image for api_server then

PR #175, opened 2021-03-05
Utilize Direct Pair Transactions

CTetford:
Revisions to binance_api_manager.py and transaction_through_bridge() to utilize direct-pair transactions to bypass the bridge currency (if they are available).

This uses market buy/sell for each direct-pair transaction to execute quickly (I'm open to comments/discussion on this if limit would be better).

I've also included a tool to analyze all available coins supported by Binance to visually see what direct-pair transactions are available to help users choose a supported coin list.

Included is a list of the largest subset of coins that can be run using only direct-pairs with no bridge.


j-waters (changes requested):
This is currently broken, with quite a few rebasing issues. I think the idea of direct pair transactions is good though.

A few things to change:

- [ ] Is `max_coin_clique.py` necessary? If it is you need to fix it and update `requirements.txt`
- [ ] Are separate `_direct_pair_buy` and `_direct_pair_sell` methods necessary? Aren't they pretty much identical to the buy and sell methods we have at the moment?
- [ ] is `supported_coin_list_direct_pair_only` needed?

CTetford:
1) Not necessary no, just useful if someone is looking to maximize the likelihood of using direct pairs when looking at a set of possible coins to use.
2) No, ideally binance_api_manager.py would have generic "market buy" "limit buy" "market sell" and "limit sell" functions that would buy/sell the maximum amount of a given origin/target. I made these mainly to separate out this functionality to avoid breaking existing code.
3) Not needed, but that list is the largest single group of coins that can all be traded with each other without requiring a bridge currency. It's included for reference.

Could you clarify what parts are broken? It seems to be running fine on my system right now. 

Feel free to improve upon this if you'd like. I'd like to contribute, but the overall code base is getting to the limits of what I'm able to help out with.

j-waters:
1. Fair enough, I'd add something to the README explaining how to use it
2. I haven't looked into it but I would have thought that they were already generic enough, maybe not though
3. Ah okay, again, maybe add a whole section to the README explaining direct pair transactions and mention this there? Also I just saw you had mentioned these things in your first comment! Sorry about that. 

Lines like [this](https://github.com/edeng23/binance-trade-bot/blob/92979281c741d6dac711f7d988048bd17400b89e/binance_trade_bot/auto_trader.py#L30) are broken - there are no longer global values for `client`, `logger`, `BRIDGE` etc. Instead do `self.client`, `self.logger`, `self.config.BRIDGE`, etc.

Not sure why it's still working for you, make sure you are fully rebased. I fetched this PR by doing `git fetch origin pull/175/head:direct_pairs` if you'd like to repeat what I did.

CTetford:
Ok I'll look through and fix them. For the buy/sell functions, I think the market sell used to change to the bridge currency will work, but I don't believe a market buy function existed before.

My intention was to stick to market buy/sell to improve the speed of the transactions and to avoid getting "stuck" if the currency value increases a little between determining what the limit is and sending the order.

My understanding is that the limit-buy was put in place to avoid losing value during the delay in switching to the bridge currency first. Since that isn't the case with direct pairs it seemed like a good compromise.

Cactii1:
Trying this...

Cactii1:
Maybe you could have it look for direct pairs by first running a call to: https://api.binance.com/api/v3/ticker/price

Cactii1:
Yeah... this is broken... Unusable as it.

CTetford:
@Cactii1 that was kind of established already...

Cactii1:
@CTetford - sure... but I did a bit of editing and had it do a sale from XMR->ETH no problem... but then there was some issues after that. I'm looking forward to being able to use this. I think it's really important.

CTetford:
Yeah it'll definitely speed up transactions for the ones that are paired.

What kind of issues did you get? (Just so I'm looking in the right direction). 

I have a feeling there's a better way of saving the ratio at the end of the transaction, but I just haven't found it yet.

Cactii1:
in auto_trader.py there was a bunch of BRIDGE references that needed "self.config." put in front of them. Then there was something with the logging that wasn't right... Just basically it was referencing things and couldn't find them.

CTetford:
@Cactii1 I've updated the references and tested it again - it's executed three trades so far without issue. Let me know if you have any other problems.

@j-waters I'll update the Readme as requested.

Cactii1:
Can you explain how the supported_coin_list_direct_pair_only works though?

For example: If I put just BTC, XMR and ETH together in that file with no alt-coins in the supported_coin_list file and run it - it gives me an error. So if I put ETH in the supported_coin_list file it will start to do something.

Cactii1:
I can confirm that this is working. It's made a direct pair trade for me.

CTetford:
I'll add a clarification to the README, the direct pair supported list is just a sample list of the maximum group of coins that are all directly tradeable with each other - you'd still have to copy/paste into your supported list.

CTetford:
@j-waters Would you be able to give me a go/no-go decision for this PR?

I'd rather not spend time maintaining conflict resolution if it isn't going to be accepted.

Thank you

Cactii1:
I've been using it constant since you made it available. Thanks @CTetford

j-waters:
@CTetford I'd probably put this PR on pause, since there are some other overhauls to trading in the works.

I do like your method, I think it could be done with slightly less modifications - like I said I don't think separate new buy/sell methods are necessary, the tool and coin list you added seems useful but should perhaps just be something we add to the wiki and not the codebase itself.

I'd definitely like something like this to be added in the future though! Plus it'd be nice to link it in with the new fee detection stuff

CTetford:
I agree, ideally there should be a single "buy" function, and a single "sell" function with input parameters that execute it as either a limit/market trade. That way there's only two functions.

Side question - the main reason I started working on this was because my script was constantly getting stuck on the bridge currency for extended periods of time. Is that issue on the roadmap for improvements?

j-waters:
I've added that in #204!

CTetford:
@j-waters That's great news! I'll wait until the overhauls you mentioned implemented and see if I can re-work the logic of this PR for the new system with minimal edits. So long as the ability to market-buy is available (not just limit-buy) it's just a matter of adding the check for pairs.

If the overhaul includes adding direct pairs, one issue I have noticed is that Binance still lists pairs in the API return that are no longer direct tradeable, e.g. BTTBTC. If a trade is executed it throws an error about "market closed" or something similar. There needs to be some kind of blacklist that gets updated for anything that is no longer available.

kiansheik:
This PR seems really useful, I will try it out soon. Currently using #204 but it doesn't have auto fee detection yet. Commenting to show support for this feature. I think with all three of these improvements, the bot will be much more streamlined and flexible.

PR #183, opened 2021-03-07
Maintain formatting and check for lint issues on CI

s0undt3ch:
* Solves a few lint issues
* disables a few other lint checks(don't want to change project's naming convention for example)
* Black the code
* Enforce all of the above for each PR


j-waters:
I definitely think enforcing a code style is a good idea, and this PR is a lot more thorough than my old one #95 so I'll close it

segersniels:
Would suggest to add the linting to the already present workflow so you can make the actual publishing of the docker image depend on the linting (splitting it up into two jobs).

.github/workflows/testing.yml
```diff
@@ -0,0 +1,37 @@
+name: CI
```
segersniels: Would suggest to add the linting to the already present workflow so you can make the actual publishing of the docker image depend on the linting (splitting it up into two jobs).

segersniels (approved):

.github/workflows/testing.yml
s0undt3ch: Then linting would only run on release?

Or you only actually prefer the CI and Release jobs to be on the same workflow?

s0undt3ch: Nevermind, will do.

s0undt3ch: Done

s0undt3ch:
How about even building the containers, to confirm it works, but only push the containers on "push" events?

s0undt3ch:
@edeng23 is this PR welcomed?
Rebasing for every change going in is not that cool, so, if not desirable, I'll just close it. I'll keep updating/rebasing otherwise.

.github/workflows/testing.yml
segersniels: Would just add a conditional to the deploy job to only run on master branch while allowing linting to run against everything.

Eg.

`if: github.ref == 'refs/heads/master'`

segersniels: Also just keep the original workflow file. No need to add a rename into this PR when it's already there.

s0undt3ch: It's no longer release only focused....

s0undt3ch: Are you sure?

s0undt3ch: > Would just add a conditional to the deploy job to only run on master branch while allowing linting to run against everything.

I had it like that at one point.
But then you don't test building the containers....

edeng23:
@s0undt3ch hey, yes, your PR is welcomed. I'll take a close look at the linting rules and merge as soon as I can. Thanks!

.github/workflows/testing.yml
segersniels: That's how it's usually done. Usually want the action to come back as positive as fast as possible without having to wait on a multi arch image to build.

s0undt3ch: I agree, as long as there's a "state" where we test if we can create the containers before releasing.

Ie, I'll be onboard with that if we push the docker containers as "dev" versions(or not publish at all) and we start versioning and tagging proper releases in git. Probably a `stable` tag on these release containers would also be suitable.

edeng23:
@s0undt3ch I'm ready to merge this PR. The only code style linting feature that I don't like is the separation of listed imports. Other than that everything looks good :)
Can you please update the .pylintrc file and rebase?
Thanks!

s0undt3ch:
That's done by the reorder python imports pre-commit hook, which also removed duplicates and separates app imports from stdlib imports from 3rd-party imports.
That also means redo most of the PR from scratch. Not impossible since some of this work is does by pre-commit and the installed hooks.

Anyway, I've moved on. If I do find time, I'll do it, but of someone wants to take over, please do.

j-waters:
I've created https://github.com/edeng23/binance-trade-bot/pull/202, which implements your suggestion @edeng23, fixes some config issues, adds some documentation, and is fully rebased

PR #184, opened 2021-03-07
Configurable heartbeat message with variables

### LINKED ISSUE #218: Option to send more output to telegram
As a way to confirm that the bot is running, it would be nice if there was an option in user.cfg to allow for more info, such as "scouting the best trade" to be sent telegram.

Brackyt:
Second PR for this as I messed up the last one.
This adds a heartbeat message with a configurable message that is sent every set amount of time (in seconds).
The message uses variables that can show different informations.
It includes the following variables:

- current_coin : Displays the current coin's symbol.
- balance: Displays the balance of the current coin.
- ratios: Displays the ratios in percentage for each pair with the current coin. With the default settings, the bot should switch coins when one of the pairs shows 0.5%. (hard to explain for me but see output below)

With this example heartbeat message: "Scouting. Current coin: {current_coin} - Balance: {balance}. Current ratios: \n{ratios}"
It will log this for example:

```
Scouting. Current coin: ETH - Balance: 0.0246. Current ratios: 
ETH to XLM. Diff: -0.23%
ETH to TRX. Diff: -0.63%
ETH to ICX. Diff: -0.79%
ETH to EOS. Diff: -0.54%
ETH to IOTA. Diff: -0.85%
ETH to ONT. Diff: -0.53%
```

If you have any other variable ideas that could be added I can add them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,8 @@ Create a .cfg file named `user.cfg` based off `.user.cfg.example`, then add your
 -   **hourToKeepScoutHistory** - Controls how many hours of scouting values are kept in the database. After the amount of time specified has passed, the information will be deleted.
 -   **scout_transaction_fee** - The transaction fee percentage. This value should be changed, for example, if you are [using BNB to pay for fees](https://www.binance.com/en/support/faq/115000583311-Using-BNB-to-Pay-for-Fees).
 -   **scout_multiplier** - Controls the value by which the difference between the current state of coin ratios and previous state of ratios is multiplied. For bigger values, the bot will wait for bigger margins to arrive before making a trade.
--   **heartbeat_duration** - The time in seconds until the heartbeat message is sent. If 0, the heartbeat message won't be used. If 3600, the heartbeat message will be sent every hour.
--   **heartbeat_message** - The message to be sent everything heartbeat_duration seconds. Configurable with variables. Possible variables are:
+-   **heartbeat_duration** - The time in minutes until the heartbeat message is sent. If 0, the heartbeat message won't be used. If 60, the heartbeat message will be sent every hour.
+-   **heartbeat_message** - The message to be sent everything heartbeat_duration minutes. Configurable with variables. Possible variables are:
     -   **current_coin**: Displays the current coin's symbol.
     -   **balance**: Displays the balance of the current coin.
     -   **ratios**: Displays the ratios in percentage for each pair with the current coin.
--- a/README.md
+++ b/README.md
@@ -57,6 +57,11 @@ Create a .cfg file named `user.cfg` based off `.user.cfg.example`, then add your
 -   **hourToKeepScoutHistory** - Controls how many hours of scouting values are kept in the database. After the amount of time specified has passed, the information will be deleted.
 -   **scout_transaction_fee** - The transaction fee percentage. This value should be changed, for example, if you are [using BNB to pay for fees](https://www.binance.com/en/support/faq/115000583311-Using-BNB-to-Pay-for-Fees).
 -   **scout_multiplier** - Controls the value by which the difference between the current state of coin ratios and previous state of ratios is multiplied. For bigger values, the bot will wait for bigger margins to arrive before making a trade.
+-   **heartbeat_duration** - The time in seconds until the heartbeat message is sent. If 0, the heartbeat message won't be used. If 3600, the heartbeat message will be sent every hour.
+-   **heartbeat_message** - The message to be sent everything heartbeat_duration seconds. Configurable with variables. Possible variables are:
+    -   **current_coin**: Displays the current coin's symbol.
+    -   **balance**: Displays the balance of the current coin.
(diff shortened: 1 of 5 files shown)


edeng23:
Thanks for sorting things out!
Can you please change the heartbeat timing granularity from seconds to minutes? I feel like most people don't want to get a notification every few seconds, but rather have the notification once every few hours. Converting hours to seconds is kind of a hassle and less elegant in my opinion.
Other than that I ran the code and it works well. 

One last thing that concerns me is that I'm not sure that auto_trader.py is the right place for the heartbeat implementation since it isn't part of the trading logic. @j-waters @rel-s What do you think?

Brackyt:
@edeng23 I'll make the change to minutes.

I think it isn't the right place too, at first it was all in crypto_trading.py but that wasn't really clean and thought that making a file just for that wasn't really a good idea.

binance_trade_bot/crypto_trading.py
```diff
@@ -32,6 +32,8 @@ def main():
     schedule.every(1).minutes.do(trader.update_values).tag("updating value history")
     schedule.every(1).minutes.do(db.prune_scout_history).tag("pruning scout history")
     schedule.every(1).hours.do(db.prune_value_history).tag("pruning value history")
+    if config.HEARTBEAT_DURATION > 0:
```
s0undt3ch, line 35:
```suggestion
    if config.HEARTBEAT_DURATION:
```
This is enough if `config.HEARTBEAT_DURATION` is an integer

s0undt3ch (changes requested):

binance_trade_bot/crypto_trading.py
Brackyt: Made it this way for code clarity

rayures:
would be nice to make this notification optional for apprise. 

in the console i would like to see every 1 minute the ratios. But i only would like to be notified via apprise about jumps.

Brackyt:
@rayures there should be many settings added to configure all notifications. Right now, doing what you asked is not even possible because the code doesn't allow to choose (code is partly there but not done correctly) to show with apprise or not

README.md
```diff
@@ -57,6 +57,11 @@ Create a .cfg file named `user.cfg` based off `.user.cfg.example`, then add your
...
 -   **scout_transaction_fee** - The transaction fee percentage. This value should be changed, for example, if you are [using BNB to pay for fees](https://www.binance.com/en/support/faq/115000583311-Using-BNB-to-Pay-for-Fees).
 -   **scout_multiplier** - Controls the value by which the difference between the current state of coin ratios and previous state of ratios is multiplied. For bigger values, the bot will wait for bigger margins to arrive before making a trade.
+-   **heartbeat_duration** - The time in minutes until the heartbeat message is sent. If 0, the heartbeat message won't be used. If 60, the heartbeat message will be sent every hour.
+-   **heartbeat_message** - The message to be sent everything heartbeat_duration minutes. Configurable with variables. Possible variables are:
+    -   **current_coin**: Displays the current coin's symbol.
+    -   **balance**: Displays the balance of the current coin.
+    -   **ratios**: Displays the ratios in percentage for each pair with the current coin.
```
rel-s, line 64: Perhaps include the curly brackets here? Just so it's clearer that these are variables. Something like:
`{current_coin}`: Displays the current coin's symbol.

binance_trade_bot/auto_trader.py
```diff
@@ -170,3 +170,50 @@ def update_values(self):
...
+
+    def get_current_ratios(self):
+        all_tickers = self.manager.get_all_market_tickers()
+
+        current_coin = self.db.get_current_coin()
+
+        current_coin_price = get_market_ticker_price_from_list(all_tickers, current_coin + self.config.BRIDGE)
+
+        if current_coin_price is None:
+            self.logger.info("Skipping scouting... current coin {0} not found".format(current_coin + self.config.BRIDGE))
```
rel-s, line 182: This whole function introduces duplicate code. I'm don't think it's a disaster, and if you don't want to deal with it we can do so in a different PR, but this log message shouldn't indicate that we are in a scouting state.

binance_trade_bot/auto_trader.py
```diff
@@ -170,3 +170,50 @@ def update_values(self):
...
+
+        heartbeat_msg = ""
+
+        for pair in self.db.get_pairs_from(current_coin):
+            if not pair.to_coin.enabled:
+                continue
+            optional_coin_price = get_market_ticker_price_from_list(all_tickers, pair.to_coin + self.config.BRIDGE)
+
+            if optional_coin_price is None:
+                self.logger.info("Skipping scouting... optional coin {0} not found".format(pair.to_coin + self.config.BRIDGE))
```
rel-s, line 193: Same goes for this guy.

mattangus:
It might be more informative to have the best ratios for each coin since last trade. Especially since it's going to happen every few minutes, the ratios at the time of the heartbeat aren't very useful.  If that is something that people would be interested in I would be happy to fork and make another PR.

PR #186, opened 2021-03-08
separate logging

Pekafu:
since api_server can be run even without the actual crypto_trading service there should be own logging for it.

### DIFF
--- a/binance_trade_bot/logger.py
+++ b/binance_trade_bot/logger.py
@@ -11,14 +11,14 @@ class Logger:
     Logger = None
     NotificationHandler = None
 
-    def __init__(self):
+    def __init__(self, loggingService = "crypto_trading"):
         # Logger setup
-        self.Logger = logging.getLogger("crypto_trader_logger")
+        self.Logger = logging.getLogger(loggingService + "_logger")
         self.Logger.setLevel(logging.DEBUG)
         formatter = logging.Formatter(
             "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
         )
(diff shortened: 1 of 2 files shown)


j-waters (approved):

binance_trade_bot/logger.py
```diff
@@ -3,22 +3,20 @@
...
 class Logger:
 
     Logger = None
     NotificationHandler = None
 
-    def __init__(self):
+    def __init__(self, loggingService = "crypto_trading"):
         # Logger setup
-        self.Logger = logging.getLogger("crypto_trader_logger")
+        self.Logger = logging.getLogger(loggingService + "_logger")
```
s0undt3ch:
```suggestion
        self.Logger = logging.getLogger(f"{loggingService}_logger")
```

binance_trade_bot/logger.py
```diff
@@ -3,22 +3,20 @@
...
         self.Logger.setLevel(logging.DEBUG)
         formatter = logging.Formatter(
             "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
         )
-        fh = logging.FileHandler(LOG_PATH)
+        # default is "logs/crypto_trading.log"
+        fh = logging.FileHandler("logs/" + loggingService + ".log")
```
s0undt3ch:
```suggestion
        fh = logging.FileHandler(f"logs/{loggingService}.log")
```

s0undt3ch (changes requested):
I think we're free to use fstrings

PR #188, opened 2021-03-08
Update docker-compose to include api user.cfg

elliotks:
Without this config available, the api container will fail due to api_key, api_secret_key settings in user.cfg being unavailable.


docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
     image: edeng23/binance-trade-bot
     container_name: binance_trader_api
     volumes:
+      - ./user.cfg:/app/user.cfg
```
segersniels: It would probably be cleaner to configure a shared volume and mount this over the entire `/app` of both containers while the bot container extracts singular files for manual configuration.

eg.
```yaml
version: "3"

services:
  crypto-trading:
    image: edeng23/binance-trade-bot
    container_name: binance_trader
    volumes:
      - foo:/app
      - ./user.cfg:/app/user.cfg
      - ./data:/app/data
      - ./logs:/app/logs
    command: python -m binance_trade_bot

  api:
    image: edeng23/binance-trade-bot
    container_name: binance_trader_api
    volumes:
      - foo:/app
    ports:
      - 5123:5123
    command: gunicorn binance_trade_bot.api_server:app -k eventlet -w 1 --threads 1 -b 0.0.0.0:5123
    depends_on:
      - crypto-trading

volumes:
  foo:
```

Doing it like this you still have all the local files that you need while just straight up sharing the `/app` between both containers. Untested though and not at home so feel free to give it some experimentation whoever wants to try.

PR #182, opened 2021-03-07
Use thread to write scout log

dv-hoang:
Now the bot takes much time to calculate the pair ratio which makes the market price is out of date when processing the transaction. The issue comes from the write scout log function. This PR to reduce the delay time.

Before
```
2021-03-07 14:49:40,024 - crypto_trader_logger - INFO - Starting calculate ratio coin: MBLUSDT 
2021-03-07 14:49:40,028 - crypto_trader_logger - INFO - optional_coin_price: <BTT>: 0.00125
2021-03-07 14:49:45,049 - crypto_trader_logger - INFO - optional_coin_price: <HOT>: 0.0030834
2021-03-07 14:49:50,074 - crypto_trader_logger - INFO - optional_coin_price: <WIN>: 0.0002113
2021-03-07 14:49:55,094 - crypto_trader_logger - INFO - optional_coin_price: <NPXS>: 0.0020162
2021-03-07 14:50:00,121 - crypto_trader_logger - INFO - optional_coin_price: <DENT>: 0.0016939
2021-03-07 14:50:05,140 - crypto_trader_logger - INFO - optional_coin_price: <KEY>: 0.007369
2021-03-07 14:50:10,159 - crypto_trader_logger - INFO - optional_coin_price: <TROY>: 0.00896
2021-03-07 14:50:15,193 - crypto_trader_logger - INFO - optional_coin_price: <DREP>: 0.008951
2021-03-07 14:50:20,212 - crypto_trader_logger - INFO - optional_coin_price: <VTHO>: 0.005925
2021-03-07 14:50:25,229 - crypto_trader_logger - INFO - Finish calculate ratio
```

After
```
2021-03-07 15:11:21,178 - crypto_trader_logger - INFO - Starting calculate ratio coin: MBLUSDT 
2021-03-07 15:11:21,182 - crypto_trader_logger - INFO - optional_coin_price: <BTT>: 0.0012462
2021-03-07 15:11:21,184 - crypto_trader_logger - INFO - optional_coin_price: <HOT>: 0.0030711
2021-03-07 15:11:21,186 - crypto_trader_logger - INFO - optional_coin_price: <WIN>: 0.00021253
2021-03-07 15:11:21,190 - crypto_trader_logger - INFO - optional_coin_price: <NPXS>: 0.0020233
2021-03-07 15:11:21,193 - crypto_trader_logger - INFO - optional_coin_price: <DENT>: 0.0017131
[...]

PR #189, opened 2021-03-08
stack scout history before writing to db

Pekafu:
Stacking seems quite efficient.
> 2021-03-08 21:04:21,133 - crypto_trader_logger - INFO - Writing scout stack to db.
> 2021-03-08 21:04:23,432 - crypto_trader_logger - INFO - All scout info written.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -31,7 +31,7 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python module names) to load,
 # usually to register additional checkers.
-load-plugins=pylint_sqlalchemy
+load-plugins=pylint_flask_sqlalchemy, pylint_sqlalchemy, pylint_flask
 
 # Pickle collected data for later comparisons.
 persistent=yes
@@ -457,7 +457,7 @@ ignore-on-opaque-inference=yes
 # List of class names for which member attributes should not be checked (useful
 # for classes with dynamically set attributes). This supports the use of
 # qualified names.
-ignored-classes=optparse.Values,thread._local,_thread._local
+ignored-classes=optparse.Values,thread._local,_thread._local,query
 
 # List of module names for which member attributes should not be checked
 # (useful for modules/projects where namespaces are manipulated during runtime
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -163,7 +163,7 @@ class AutoTrader:
             ) - pair.ratio
 
 
-        if(len(scout_stack) > 0):
+        if scout_stack:
             self.db.log_scout_stack(scout_stack)
             scout_stack.clear()
 
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -9,7 +9,7 @@ from .binance_api_manager import BinanceAPIManager
 from .config import Config
 from .database import Database
 from .logger import Logger
-from .models import Coin, CoinValue, Pair
+from .models import Pair, Coin, CoinValue, ScoutHistory
 from .utils import get_market_ticker_price_from_list
 
 
@@ -135,7 +135,7 @@ class AutoTrader:
             return
 
         ratio_dict: Dict[Pair, float] = {}
-
+        scout_stack = []
         for pair in self.db.get_pairs_from(current_coin):
             if not pair.to_coin.enabled:
                 continue
@@ -147,7 +147,8 @@ class AutoTrader:
                 )
                 continue
 
-            self.db.log_scout(pair, pair.ratio, current_coin_price, optional_coin_price)
+            
+            scout_stack.append(ScoutHistory(pair, pair.ratio, current_coin_price, optional_coin_price))
 
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = current_coin_price / optional_coin_price
@@ -160,6 +161,11 @@ class AutoTrader:
             ratio_dict[pair] = (
                 coin_opt_coin_ratio - transaction_fee * self.config.SCOUT_MULTIPLIER * coin_opt_coin_ratio
             ) - pair.ratio
+
+
+        if(len(scout_stack) > 0):
+            self.db.log_scout_stack(scout_stack)
+            scout_stack.clear()
 
         # keep only ratios bigger than zero
         ratio_dict = {k: v for k, v in ratio_dict.items() if v > 0}
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -9,7 +9,7 @@ from .binance_api_manager import BinanceAPIManager
 from .config import Config
 from .database import Database
 from .logger import Logger
-from .models import Pair, Coin, CoinValue, ScoutHistory
+from .models import Coin, CoinValue, Pair, ScoutHistory
 from .utils import get_market_ticker_price_from_list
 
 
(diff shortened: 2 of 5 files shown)


Pekafu:
Relates to #182

binance_trade_bot/auto_trader.py
```diff
@@ -132,14 +132,20 @@ def scout(self):
...
+            scout_stack.append(ScoutHistory(pair, pair.ratio, current_coin_price, optional_coin_price))
 
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = current_coin_price / optional_coin_price
 
             # save ratio so we can pick the best option, not necessarily the first
             ratio_dict[pair] = (coin_opt_coin_ratio - self.config.SCOUT_TRANSACTION_FEE * self.config.SCOUT_MULTIPLIER * coin_opt_coin_ratio) - pair.ratio
 
+
+        if(len(scout_stack) > 0):
```
s0undt3ch:
```suggestion
        if scout_stack:
```
This is enough.

s0undt3ch (changes requested):

binance_trade_bot/database.py
```diff
@@ -179,6 +192,19 @@ def create_database(self):
     def start_trade_log(self, from_coin: Coin, to_coin: Coin, selling: bool):
         return TradeLog(self, from_coin, to_coin, selling)
 
+    def send_bulk_update(self, models):
+        if not self.socketio_connect():
+            return
+        
+        bulk_data = []
```
j-waters:
```suggestion
        bulk_data = list(model.info() for model in models)
```

Just how I would do it feel free to ignore!

binance_trade_bot/database.py
```diff
@@ -121,6 +121,19 @@ def get_pairs_from(self, from_coin: Union[Coin, str]):
...
             return pairs
 
+
+    def log_scout_stack(self, sh_stack):
+        session: Session
+        with self.db_session() as session:
+            merged_shs = []
+            for sh_temp in sh_stack:
+                pair = session.merge(sh_temp.pair)
+                sh = ScoutHistory(pair, sh_temp.target_ratio, sh_temp.current_coin_price, sh_temp.other_coin_price)
```
j-waters: Is there a reason we're creating a new object here? I'm sure there is because SQLAlchemy can sometimes be tricky but if there's another way we should do that.

```suggestion
                session.add(sh_temp)
```

Does that not work? What if you merged it first and then did that?

binance_trade_bot/database.py
```diff
@@ -179,6 +192,19 @@ def create_database(self):
...
 
+    def send_bulk_update(self, models):
+        if not self.socketio_connect():
+            return
+        
+        bulk_data = []
+        for model in models:
+            bulk_data.append(model.info()) 
+
+        self.socketio_client.emit('bulk_save_objects', {
```
j-waters:
```suggestion
        self.socketio_client.emit('update_bulk', {
```

To keep in line with the other emitter

binance_trade_bot/database.py
```diff
@@ -121,6 +121,19 @@ def get_pairs_from(self, from_coin: Union[Coin, str]):
             pairs: List[Pair] = session.query(Pair).filter(Pair.from_coin == from_coin)
             return pairs
 
+
+    def log_scout_stack(self, sh_stack):
```
j-waters:
```suggestion
    def log_scout_stack(self, sh_stack: List[ScoutHistory]):
```

j-waters (changes requested):
This is a good idea, thanks for implementing.

My only thought is if we could come up with a more general solution, but that's probably more effort than it'd be worth

Also do you have a ballpark on how much time this saves? Just so we don't overcomplicate things for no reason!

binance_trade_bot/database.py
Pekafu: this looks lovely, will definitely change.

binance_trade_bot/database.py
Pekafu: I was basically mimicing the old implementation and since I'm not exactly sure what that session.merge() does I left it as it is. Checking through that pair might not be needed, since ScoutHistory shouldn't exist without a pair since we ain't updating pairs on the fly.

Pekafu: > 
> 
> This is a good idea, thanks for implementing.
> 
> My only thought is if we could come up with a more general solution, but that's probably more effort than it'd be worth
> 
> Also do you have a ballpark on how much time this saves? Just so we don't overcomplicate things for no reason!

I tested it with stacking and running stack in a loop through the old log_scout() and it took more than a minute, now it goes below 3sec, still it should be possible to get it below 1sec.

j-waters: `session.merge` takes an object that was created in another database session, refreshes it, and links it with the current session. We have to do this with pairs since they're created in another session.

You could do this:
```suggestion
                sh = session.merge(sh_temp)
```

and it might automatically merge the pair for you, I'm not sure. If that doesn't work, try:
```suggestion
                pair = session.merge(sh_temp.pair)
                sh_temp.pair = pair
```
Something like that? I can test it myself later if needs be

j-waters: Also don't know why it took more than a minute originally! That's a good speed increase

Pekafu: I don't have time to test in within next couple of days, so go ahead =)

rayures:
forked and merged this one. 
Works great. time between jump notification to buy order went from 50-ish seconds to 5 seconds.

alienfrenZyNo1:
I've implemented this and being running it the last few days. It hasn't broken anything. I have noticed less buy limit misses but it still happens. I like this PR though, understandable and focused on one thing!

binance_trade_bot/database.py
rel-s: Is this resolved? Does @j-waters modification work?

j-waters: I never ended up testing it, I can get round to it though

rayures: running it for a couple of days. works fine.

.pylintrc
```diff
@@ -457,7 +457,7 @@ ignore-on-opaque-inference=yes
 # List of class names for which member attributes should not be checked (useful
 # for classes with dynamically set attributes). This supports the use of
 # qualified names.
-ignored-classes=optparse.Values,thread._local,_thread._local
+ignored-classes=optparse.Values,thread._local,_thread._local,query
```
Pekafu, line 460: couldn't get pre-commit hooks to work without these.

binance_trade_bot/database.py
```diff
@@ -121,13 +121,16 @@ def get_pairs_from(self, from_coin: Union[Coin, str]):
...
             return pairs
 
-    def log_scout(
-        self,
-        pair: Pair,
-        target_ratio: float,
-        current_coin_price: float,
-        other_coin_price: float,
-    ):
+    def log_scout_stack(self, sh_stack: List[ScoutHistory]):
```
Pekafu, line 124: with this version it seems to be working within 3secs, I don't currently have a lot of pairs though
> 48:20,989 - crypto_trading_logger - INFO - started stacking
> 48:23,293 - crypto_trading_logger - INFO - stacking rdy
> 48:28,633 - crypto_trading_logger - INFO - started stacking
> 48:30,941 - crypto_trading_logger - INFO - stacking rdy                           
> 48:36,271 - crypto_trading_logger - INFO - started stacking
> 48:38,579 - crypto_trading_logger - INFO - stacking rdy

rel-s (approved):

PR #190, opened 2021-03-09
 Turn bot into an installable package which can be cythonized 

s0undt3ch:
The gain here is to squeeze as much performance from the bot as possible by running the bot under C(ithonized) instead of python


rel-s:
Closing for now, due to the combination of inactivity and "draft" status. Do comment if you want this reopened.

PR #191, opened 2021-03-09
Don't traceback if `user.cfg` does not exist / De-duplicate coins

s0undt3ch:
De-duplicate coins list when loading from file


rel-s (approved):
Will test this soon!

PR #195, opened 2021-03-11
Update trading logic to support cancel timeout order

dv-hoang:
Features:
- [x] Sell at calculated price instead of market price to avoid losing the coin
- [x] Handle cancel sell/buy order when timeout and have no filled
- [x] Handle cancel sell.buy order when timeout and partially filled

Update:
- Remove `bridge_scout` which makes the bot buy another coin after canceling the sell order. It will make you lost the coin

### DIFF
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -20,22 +20,34 @@ class AutoTrader:
     def initialize(self):
         self.initialize_trade_thresholds()
 
-    def transaction_through_bridge(self, pair: Pair, all_tickers):
+    def transaction_through_bridge(self, pair: Pair, all_tickers: AllTickers):
         """
         Jump from the source coin to the destination coin through bridge coin
         """
-        if self.manager.sell_alt(pair.from_coin, self.config.BRIDGE) is None:
+        can_sell = False
+        balance = self.manager.get_currency_balance(pair.from_coin.symbol)
+        from_coin_price = all_tickers.get_price(pair.from_coin + self.config.BRIDGE)
+
+        if balance and balance * from_coin_price > self.manager.get_min_notional(pair.from_coin, self.config.BRIDGE):
+            can_sell = True
+        else:
+            self.logger.info("Skipping sell")
+
+        if can_sell and self.manager.sell_alt(pair.from_coin, self.config.BRIDGE, all_tickers) is None:
             self.logger.info("Couldn't sell, going back to scouting mode...")
             return None
 
         result = self.manager.buy_alt(pair.to_coin, self.config.BRIDGE, all_tickers)
+
         if result is not None:
+            self.db.set_current_coin(pair.to_coin)
             self.update_trade_threshold(pair.to_coin, float(result["price"]), all_tickers)
             return result
+
         self.logger.info("Couldn't buy, going back to scouting mode...")
         return None
 
-    def update_trade_threshold(self, coin: Coin, coin_price: float, all_tickers):
+    def update_trade_threshold(self, coin: Coin, coin_price: float, all_tickers: AllTickers):
         """
         Update all the coins with the threshold of buying the current held coin
         """
--- a/binance_trade_bot/config.py
+++ b/binance_trade_bot/config.py
@@ -8,7 +8,7 @@ CFG_FL_NAME = "user.cfg"
 USER_CFG_SECTION = "binance_user_config"
 
 
-class Config:  # pylint: disable=too-few-public-methods
+class Config:  # pylint: disable=too-few-public-methods,too-many-instance-attributes
     def __init__(self):
         # Init config
         config = configparser.ConfigParser()
@@ -19,6 +19,8 @@ class Config:  # pylint: disable=too-few-public-methods
             "hourToKeepScoutHistory": "1",
             "tld": "com",
             "strategy": "default",
+            "sell_timeout": "0",
+            "buy_timeout": "0",
         }
 
         if not os.path.exists(CFG_FL_NAME):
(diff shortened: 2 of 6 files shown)


binance_trade_bot/auto_trader.py
```diff
@@ -132,23 +147,33 @@ def scout(self):
                 self.logger.info("Skipping scouting... optional coin {0} not found".format(pair.to_coin + self.config.BRIDGE))
                 continue
 
-            self.db.log_scout(pair, pair.ratio, current_coin_price, optional_coin_price)
+            hitory = ScoutHistory(pair, pair.ratio, current_coin_price, optional_coin_price)
+            scout_stack.append(hitory)
```
Pekafu: probably don't need additional parameter, but if you want to use it should be named history

binance_trade_bot/auto_trader.py
```diff
@@ -19,17 +21,27 @@ def __init__(self, binance_manager: BinanceAPIManager, database: Database, logge
...
-        result = None
-        while result is None:
-            result = self.manager.buy_alt(pair.to_coin, self.config.BRIDGE, all_tickers)
+        result = self.manager.buy_alt(pair.to_coin, self.config.BRIDGE, to_coin_price)
+
+        if result is None:
+            return None
 
         self.db.set_current_coin(pair.to_coin)
         self.update_trade_threshold(float(result[u'price']), all_tickers)
```
Pekafu: all_tickers is undefined here.

j-waters:
`min_notional` can and should be fetched using the API, see #137 https://github.com/edeng23/binance-trade-bot/blob/85777d3af3e8f4ac14cc1b5cc7d5bd8f8cbccc51/binance_trade_bot/binance_api_manager.py#L71

binance_trade_bot/auto_trader.py
```diff
@@ -132,24 +149,34 @@ def scout(self):
...
             best_pair = max(ratio_dict, key=ratio_dict.get)
-            self.logger.info('Will be jumping from {0} to {1}'.format(
-                current_coin, best_pair.to_coin_id))
+            self.logger.info('Will be jumping from {0} to {1}'.format(current_coin, best_pair.to_coin_id))
+
+            optional_coin_price = get_market_ticker_price_from_list(all_tickers, best_pair.to_coin + self.config.BRIDGE)
             self.transaction_through_bridge(best_pair, all_tickers)
 
+        # sleep to avoid reach API request limit
+        time.sleep(5)
```
j-waters: Maybe instead of sleeping we just catch the error we get when we hit the API request limit? We already have a system for adding a delay between scouts (the issue is that at the moment scouting usually takes longer than the 5 second delay)

Cactii1:
Should be done differently... Client side instead of single ticker requests.

https://binance-docs.github.io/apidocs/spot/en/#symbol-price-ticker

**If the symbol is not sent, prices for all symbols will be returned in an array.**

binance_trade_bot/auto_trader.py
```diff
@@ -132,24 +149,34 @@ def scout(self):
...
+            diff = (coin_opt_coin_ratio - self.config.SCOUT_TRANSACTION_FEE * self.config.SCOUT_MULTIPLIER * coin_opt_coin_ratio) - pair.ratio
+
+            # keep only ratios bigger than min rate
+            if ((diff / pair.ratio) * 100) > float(self.config.MIN_SCOUT_RATE):
+                ratio_dict[pair] = diff
 
-        # keep only ratios bigger than zero
-        ratio_dict = {k: v for k, v in ratio_dict.items() if v > 0}
+        if scout_stack:
+            thread = threading.Thread(group=None, target=self.db.log_scout_stack, name=None, args=(scout_stack,))
```
rel-s: Didn't test the timing here, but for the sake of efficiency wouldn't it be better to start the thread before we actually get the values from Binance and pass data to it using a queue or something?

binance_trade_bot/database.py
```diff
@@ -129,6 +129,30 @@ def log_scout(self, pair: Pair, target_ratio: float, current_coin_price: float,
...
+            self.send_bulk_update(merged_shs)
+
+    def send_bulk_update(self, models):
+        if not self.socketio_connect():
+            return
+
+        bulk_data = list(model.info() for model in models)
+
+        self.socketio_client.emit('update_bulk', {
+            "table": models[0]._tablename__,
```
vizzito: BUG: typo bug: an _ is missing
change `_tablename__, by __tablename__, `
I already checked all and work very good so far.

dv-hoang: Thank you, I'm using the https://github.com/edeng23/binance-trade-bot/pull/189 and did not realize that bug

binance_trade_bot/auto_trader.py
dv-hoang: thanks @j-waters, I have removed it

binance_trade_bot/auto_trader.py
j-waters: Also personally I'd move starting the thread into the database class, for better abstraction

dv-hoang: thanks @rel-s and @j-waters, I remove the change for scout history first. May have other PR for that. This PR mostly to updating trading logic to support cancel timeout orders. Don't want it stuck by other implements

edeng23:
@kydrenw I truly appreciate the work put in to this. This certainly solves the most complained about behaviour of the bot. I'm going to start running some tests on the code. In the mean time, please rebase so that the code can be merged. 
Other than that, I think that documentation of this new behaviour should be added to the readme file. I don't mind doing that myself after testing.

binance_trade_bot/config.py
```diff
@@ -19,6 +19,8 @@ def __init__(self):
             "hourToKeepScoutHistory": "1",
             "tld": "com",
             "strategy": "default",
+            "sell_timeout": "0",
```
edeng23, line 22: Please add these to user.cfg.example

edeng23:
Ok, so after running a single test i realised that the code doesn't work the way I expected it to. 
After cancelling a buy attempt, the bot thinks that the buy attempt was successful and continues from the wrong coin. 
Moreover, it started to concatenate coin names:
`crypto_trading_logger - INFO - Will be jumping from <HOT> to DOGEHOTUSDT`

dv-hoang:
this PR is out of date due to the project structure was updated.

dv-hoang:
@edeng23 I have updated the PR to handle cancel the order for both default and multiple coins strategies.

tonysala (approved):

rel-s (approved):

PR #196, opened 2021-03-11
Added basic backtesting

j-waters:
It's rather slow and doesn't have any pretty logging - I'll get to that later.

I've created a mock Binance manager class, that simulates buying and selling, and can look up the historical ticker prices of coins. The API only allows you to get past data in 1 minute intervals, so the backtesting code runs the scout function in a loop without a delay, and after every scout an internal datetime value is incremented by 1 minute.

I also added some caching, <s>because as I said, it's rather slow. We can't do `get_all_market_tickers`, so each ticker has to be fetched individually, meaning that each run of the scout method takes about 3 seconds. It's not the end of the world, but it means testing a new config tweak isn't exactly instant. The cache does mean that once you've run the backtesting once, subsequent runs are a lot quicker. I added a script to fetch and cache data as quickly as possible, but then we run into Binance's throttling of 1200 requests per minute. Maybe we can sort something out where someone generates a sizeable amount of backtesting data and other people can just download it. Or maybe we get the data from a source other than Binance - it shouldn't make too much of a difference.</s>

PR #197, opened 2021-03-11
Revert renaming of `alt_coin` and `crypto_coin` variables

j-waters:
I appreciate what this commit was trying to do, and maybe the names could be improved, but the names as they stand don't actually make sense. I was working on #196 and kept getting confused. There is no target or origin, just the two halves of the trading pair.

The original name made sense because traditionally, a trading pair will have a more established 'crypto' coin, and a less popular 'alt' coin. It is still confusing since all the coins are crypto coins, but I think it makes more sense than `origin` and `target`, since half the time they're the wrong way round anyway.

### DIFF
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -54,19 +54,19 @@ class BinanceAPIManager:
                 attempts += 1
         return None
 
-    def get_alt_tick(self, origin_symbol: str, target_symbol: str):
+    def get_alt_tick(self, alt_symbol: str, crypto_symbol: str):
         step_size = next(
-            _filter['stepSize'] for _filter in self.BinanceClient.get_symbol_info(origin_symbol + target_symbol)['filters']
+            _filter['stepSize'] for _filter in self.BinanceClient.get_symbol_info(alt_symbol + crypto_symbol)['filters']
             if _filter['filterType'] == 'LOT_SIZE')
         if step_size.find('1') == 0:
             return 1 - step_size.find('.')
         else:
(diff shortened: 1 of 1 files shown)


edeng23:
I completely agree with you on the matter, I also got confused when playing around with the "buy" function. That being said, I don't think that "crypto" and "alt" are indicative enough. Why not use "origin" and "target"  in the logical order?

j-waters:
That would make slightly more sense, although I do think the emphasis should really be on the trading pair rather than the individual coins. [This article](https://www.gemini.com/cryptopedia/what-are-cryptocurrency-pairs) uses the term 'base', so maybe we say `alt_coin` and `base_coin`? Or `target_coin` and `base_coin`? I guess `base` is similar to origin now that I think about it.

Maybe we could change it so that the buy and sell methods accept `Pair`s instead.

So instead of `buy(Coin('XLM'),Coin( 'USDT'))` and `sell(Coin('USDT'), Coin('XLM'))` it would be `buy(Pair('XLM', 'USDT')`, etc.

Then again maybe it doesn't really matter! This isn't exactly high priority just something I noticed while developing

edeng23:
I think that for readability, giving the symbols names is important. I find that when reading the order_quantity calculation, which is quite complicated as is, having the idea of the coin you're starting from and coin you want to end up with helps. 
Using Pairs sounds like the standard way to go, addressing the coins as "from_coin" and "to_coin".

rel-s:
I think we should close this and perhaps open a new issue discussing future namings - do comment if you think we should keep this PR open.

PR #198, opened 2021-03-12
Automatically get transaction fees

### LINKED ISSUE #165: optimum scout_transaction_fee
I don't understand the "scout_transaction_fee", what value should be set?

Mine is set at 0.001, and I am noticing loss in overall coins amount..

I don't have BNB in my account so the fee is 0.1% 

Which value should be set in scout_transaction_fee?

j-waters:
Closes #165

### DIFF
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -152,8 +152,9 @@ class AutoTrader:
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = current_coin_price / optional_coin_price
 
-            transaction_fee = self.manager.get_trade_fees()[pair.from_coin + self.config.BRIDGE] + \
-                              self.manager.get_trade_fees()[pair.to_coin + self.config.BRIDGE]
+            transaction_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True) + self.manager.get_fee(
+                pair.to_coin, self.config.BRIDGE, False
+            )
 
             # save ratio so we can pick the best option, not necessarily the first
             ratio_dict[pair] = (
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -19,6 +19,7 @@ class AutoTrader:
         self.db = database
         self.logger = logger
         self.config = config
+        self.trade_fees = self.manager.get_trade_fees()
 
     def transaction_through_bridge(self, pair: Pair, all_tickers):
         """
@@ -152,10 +153,14 @@ class AutoTrader:
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = current_coin_price / optional_coin_price
 
+            transaction_fee = (
+                self.trade_fees[pair.from_coin + self.config.BRIDGE]
+                + self.trade_fees[pair.to_coin + self.config.BRIDGE]
+            )
+
             # save ratio so we can pick the best option, not necessarily the first
             ratio_dict[pair] = (
-                coin_opt_coin_ratio
-                - self.config.SCOUT_TRANSACTION_FEE * self.config.SCOUT_MULTIPLIER * coin_opt_coin_ratio
+                coin_opt_coin_ratio - transaction_fee * self.config.SCOUT_MULTIPLIER * coin_opt_coin_ratio
             ) - pair.ratio
 
         # keep only ratios bigger than zero
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -19,7 +19,6 @@ class AutoTrader:
         self.db = database
         self.logger = logger
         self.config = config
-        self.trade_fees = self.manager.get_trade_fees()
 
     def transaction_through_bridge(self, pair: Pair, all_tickers):
         """
@@ -153,10 +152,8 @@ class AutoTrader:
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = current_coin_price / optional_coin_price
 
-            transaction_fee = (
-                self.trade_fees[pair.from_coin + self.config.BRIDGE]
-                + self.trade_fees[pair.to_coin + self.config.BRIDGE]
-            )
+            transaction_fee = self.manager.get_trade_fees()[pair.from_coin + self.config.BRIDGE] + \
+                              self.manager.get_trade_fees()[pair.to_coin + self.config.BRIDGE]
 
             # save ratio so we can pick the best option, not necessarily the first
             ratio_dict[pair] = (
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -1,5 +1,6 @@
 import math
 import time
+from typing import Dict
 
 from binance.client import Client
 from binance.exceptions import BinanceAPIException
@@ -20,6 +21,9 @@ class BinanceAPIManager:
         )
         self.db = db
         self.logger = logger
+
+    def get_trade_fees(self) -> Dict[str, float]:
+        return {ticker["symbol"]: ticker["taker"] for ticker in self.binance_client.get_trade_fee()["tradeFee"]}
 
     def get_all_market_tickers(self):
         """
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -4,7 +4,7 @@ from typing import Dict
 
 from binance.client import Client
 from binance.exceptions import BinanceAPIException
-from cachetools import cached, TTLCache
+from cachetools import TTLCache, cached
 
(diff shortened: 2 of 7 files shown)


rel-s:
Looks great, testing now

rel-s (approved):
Code looks great, I wonder if we should implement another timer to update the fees (right now they are only updated on AutoTrade init). Do they actually ever change? Probably no need to implement in this PR, what do you think @edeng23 ?

j-waters:
@rel-s good point, the only reason they would change is if you were trading enough to [move into another fee bracket](https://www.binance.com/en/fee/schedule) - probably unlikely for our users.

An alternative, which I experimented with in #137, would be to use a caching library like [cachetools](https://pypi.org/project/cachetools/). Calls to `get_alt_tick` and `get_trade_fees`, which are unlikely to ever change but might one day, could simply be cached with a TTL of a day or so.

j-waters:
Pretty simple to implement. I can revert if needs be

rel-s:
Seems like a good idea to me. Could you pin the new dependency though? I think it's a good idea to start doing that, and I plan on pinning all of our current ones...

j-waters:
@rel-s done, yeah it's a good idea

rel-s (approved):
Approved, testing again...

BelphegorPrime:
> @rel-s good point, the only reason they would change is if you were trading enough to [move into another fee bracket](https://www.binance.com/en/fee/schedule) - probably unlikely for our users.


it is also possible when you pay fees with BNB and run out of it that fee percentages are changing.

rel-s:
> it is also possible when you pay fees with BNB and run out of it that fee percentages are changing.

I couldn't find any info in the official API the merely having BNB (which is the requirement for paying a lower fee, right?) will change the result of the API call.
https://binance-docs.github.io/apidocs/spot/en/#asset-detail-user_data
Why do you think it will @BelphegorPrime?

BelphegorPrime:
you can see here at the bottom on point [4]( https://www.binance.com/en/support/faq/115000583311-Using-BNB-to-Pay-for-Fees) that when you don´t have enough BNB to pay for trading fees, it will do as normal in the fee schedule.
question is, does the api return the a fee as in the fee schedule or with the BNB bonus? I just assume in the following with the BNB bonus, because I cpulnd´t find something in the api definition.
If it is with the BNB bonus it can change between single trades.
For example you have enough BNB to pay the fees for the sell but than you don´t have enough to buy the target coin. which can be result in less pr even negative profit.

EDIT:
there is an api to check if BNB "burn" is activated https://binance-docs.github.io/apidocs/spot/en/#get-bnb-burn-status-user_data

j-waters:
Interesting, yeah we can use `get_bnb_burn_spot_margin`, and then multiply the fees by 0.75? Because I'm using BNB for fees, I have enough, and it's saying that my fees are `0.001`.

It doesn't matter too much if we overestimate the fees, since it will just mean the bot is more conservative.

Edit: Done

BelphegorPrime:
> For example you have enough BNB to pay the fees for the sell but than you don´t have enough to buy the target coin. which can be result in less or even negative profit.

But this problem remains.

When BNB burn is enabled but you don`t have enough BNB to pay the fees, you are not getting the bonus for upcoming trades until you have BNB again.

Check should be something like
if bnbBurnActive == true && doesHaveEnoughBnb()
then transaction_fee *= 0.75

BelphegorPrime:
When BNB burn is active does self.BinanceClient.get_trade_fee() return the fees already with the BNB bonus of 25% off or is it the default fee schedule traiding fee?

j-waters:
Shouldn't be too difficult to implement, I'll look into it tomorrow. We might be over complicating things when it wouldn't make much difference to just leave the raw things but as long as we're not adding too many extra API calls we may as well do this.

rel-s:
Agreed - solving the BNB fee issue automatically is probably the way to go here, especially since we removed the option to manually set the transaction fee.

rel-s:
Agreed - solving the BNB fee issue automatically is probably the way to go here, especially since we removed the option to manually set the transaction fee.

j-waters:
Okay I've added logic to check if you have enough BNB. The slight limitation is that the BNB has to be in your spot account - I keep mine in an earn account and there's no way of getting that unfortunately. Not the end of the world though since it will just mean that the bot is ever so slightly more conservative.

binance_trade_bot/auto_trader.py
```diff
@@ -70,7 +70,7 @@ def initialize_trade_thresholds(self):
             for pair in session.query(Pair).filter(Pair.ratio.is_(None)).all():
                 if not pair.from_coin.enabled or not pair.to_coin.enabled:
                     continue
-                self.logger.info(f"Initializing {pair.from_coin} vs {pair.to_coin}", False)
```
rel-s: I think this isn't rebased properly, why would this change?

binance_trade_bot/config.py
```diff
@@ -24,8 +23,7 @@ def __init__(self):
         if not os.path.exists(CFG_FL_NAME):
             print("No configuration file (user.cfg) found! See README. Assuming default config...")
             config[USER_CFG_SECTION] = {}
-        else:
```
rel-s: This as well, we added this in a recent PR, this PR shouldn't break it.

rel-s:
@j-waters I skimmed through the code and it seems like it wasn't rebased properly. Could you fix this so we can re-test?

j-waters:
Ah yep sorry, all fixed

binance_trade_bot/auto_trader.py
```diff
@@ -152,10 +152,16 @@ def scout(self):
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = current_coin_price / optional_coin_price
 
+            transaction_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True) + self.manager.get_fee(
+                pair.to_coin, self.config.BRIDGE, False
+            )
+
+            if self.manager.get_using_bnb_for_fees():
```
rel-s: I think this code is already inside `get_fee()`, since it will return `base_fee * 0.75` if we are using BNB; why do we need to take this into account again?

binance_trade_bot/binance_api_manager.py
```diff
@@ -97,6 +125,17 @@ def wait_for_order(self, origin_symbol, target_symbol, order_id):
     def buy_alt(self, origin_coin: Coin, target_coin: Coin, all_tickers):
         return self.retry(self._buy_alt, origin_coin, target_coin, all_tickers)
 
+    def _buy_quantity(
```
rel-s, line 128: Not entirely sure why we need this new function.. How is this related to automatic transaction fee fetching?

binance_trade_bot/binance_api_manager.py
```diff
@@ -143,6 +180,12 @@ def _buy_alt(self, origin_coin: Coin, target_coin: Coin, all_tickers):
     def sell_alt(self, origin_coin: Coin, target_coin: Coin):
         return self.retry(self._sell_alt, origin_coin, target_coin)
 
+    def _sell_quantity(self, origin_symbol: str, target_symbol: str, origin_balance: float = None):
```
rel-s, line 183: Same comment here as in `_buy_quantity`

rel-s (commented):
I have a few questions about this PR, see my review. Anyways, I'm already starting to test this in my environment :)

binance_trade_bot/auto_trader.py
j-waters: Ah yep I thought I had removed that but apparently not thanks

binance_trade_bot/binance_api_manager.py
j-waters: Okay, so to determine if we have enough BNB for it to be used, we need to calculate how much the fee will be. That's done [here](https://github.com/edeng23/binance-trade-bot/pull/198/files#diff-65cde28df70e6f927481325a29395037472fa5a4b7700e7e2fdc3a411891c4c9R39). The BNB amount will be the amount we're trading, multiplied by the base fee, multiplied by 0.75 (the discount), multiplied by the conversion rate between the coin we're trading and BNB. We can then compare that BNB amount with our balance.

Getting the amount that we're trading is the reason I made new functions, so that we're not duplicating any logic, and so that changes to the formula don't have to be made in multiple places. Also, I've done a similar thing in #196.

binance_trade_bot/binance_api_manager.py
j-waters: Same reason as above

rel-s (approved):
Looks great, testing...

rel-s:
Seems to work perfectly on my setup! One last thing for this PR - we should probably write something in the `README` to let users know that if they have BNB, the reduced fee will be used automatically. Perhaps @edeng23 has an idea?

edeng23 (approved):

PR #199, opened 2021-03-12
Recover from being stuck in bridge coin

johannes-peeters:


j-waters:
Very big pull request so I haven't looked at it too deeply, the first thing I'd say is that there needs to be something here to migrate the user's database schema, since you're adding and removing columns.

Could you give more of a description of all the changes? I think there are some good ideas here but the key thing that attracted me to this bot was its simplicity, and this adds quite a bit of complexity. Not necessarily a bad thing, I'm just being a bit slow trying to work out what's going on.

johannes-peeters:
Yeah sorry for all the changes. I wanted to get you a preview in how to handle BRIDGE -> coin. I'll provide more context.

j-waters:
Ah thanks, I'll look it more detail later!

johannes-peeters:
Looks like not all my changes had been pushed before. Done now

abanchev (commented):
Yeah, good luck doing it this way
if you reduce the amount of money the bot is trading with, it will stop trading until the market moves up at least the percentage you took out

johannes-peeters:
> Yeah, good luck doing it this way
> if you reduce the amount of money the bot is trading with, it will stop trading until the market moves up at least the percentage you took out

The amount of money the bot is trading with isn't reducing. If in bridge coin, the bot will wait for a new trade comparing previous trade with the current expected quantity - which I believe is a fair thing to do

abanchev:
> > Yeah, good luck doing it this way
> > if you reduce the amount of money the bot is trading with, it will stop trading until the market moves up at least the percentage you took out
> 
> The amount of money the bot is trading with isn't reducing. If in bridge coin, the bot will wait for a new trade comparing previous trade with the current expected quantity - which I believe is a fair thing to do

The point is that the amount can change by actions taken by the user - if i see that the market is going down, i can sell half the coin we hold and take the money into a different stable coin (so no longer accessible to the bot), or even as take profit if the bot is doing well the user may decide to take a percentage out as cash to spend.

johannes-peeters:
> > > Yeah, good luck doing it this way
> > > if you reduce the amount of money the bot is trading with, it will stop trading until the market moves up at least the percentage you took out
> > 
> > 
> > The amount of money the bot is trading with isn't reducing. If in bridge coin, the bot will wait for a new trade comparing previous trade with the current expected quantity - which I believe is a fair thing to do
> 
> The point is that the amount can change by actions taken by the user - if i see that the market is going down, i can sell half the coin we hold and take the money into a different stable coin (so no longer accessible to the bot), or even as take profit if the bot is doing well the user may decide to take a percentage out as cash to spend.

Absolutely. The premise here is that the bot is managing the account. Manual manipulations will indeed throw it off. I am working on a version allowing manual manipulations, while keeping the premises to only increase in coin amount (in that case based on the at the latest trade price * the then current holding)

abanchev:
All these things are solved by using the coin ratio, it is a price ratio, so the bot does not care how much you are holding or did hold in the past. All issues regarding lost coins(bought less than last time) are caused by delays between when the ratio was calculated(pre sale) and when the transaction happened(N seconds later), if you instead cache the price ratio for the coin that was sold (at the time of sale, not at the current time) this will make sure any purchases are profitable for the user.

Don't get me wrong, the idea of comparing absolute amount is good at first glance, but i already tested it, and in the end it was very problematic because in real life the account is never 100% managed by the bot - the simplest example is when you sell part of the current coin to buy some more BNB to reduce your fees.

j-waters:
I implemented a different bridge scouting method in #137, which uses ratios

johannes-peeters:
> All these things are solved by using the coin ratio, it is a price ratio, so the bot does not care how much you are holding or did hold in the past. All issues regarding lost coins(bought less than last time) are caused by delays between when the ratio was calculated(pre sale) and when the transaction happened(N seconds later), if you instead cache the price ratio for the coin that was sold (at the time of sale, not at the current time) this will make sure any purchases are profitable for the user.
> 
> Don't get me wrong, the idea of comparing absolute amount is good at first glance, but i already tested it, and in the end it was very problematic because in real life the account is never 100% managed by the bot - the simplest example is when you sell part of the current coin to buy some more BNB to reduce your fees.

> I implemented a different bridge scouting method in #137, which uses ratios

Great I'll switch to #137 and test. Assume it works fine my PR can be discarded.

Issue #181, opened 2021-03-07
Pin dependencies!


Knifa:
Please pin your dependencies in `requirements.txt` so they don't change between builds! Could be related to #178 e.g.

rel-s:
Agreed, we should do this...

j-waters:
Would it also be a good idea to encourage people to use `virtualenv` instead of installing dependencies globally?

rel-s:
I think we can leave that up to the user, I don't mind putting it in the `README` but I also don't want to clutter it.

rel-s:
Done!

PR #200, opened 2021-03-13
Pin requirements

Knifa:
For #181.

I just froze whatever works for me right now (installed a few days ago).

PR #201, opened 2021-03-13
Pinned all Dependencies

rel-s:
Tested on a working setup, these should all be latest versions.


rel-s:
Should resolve #181.

rel-s:
Closing in favor of #200.

PR #205, opened 2021-03-14
Add initial implementation of websockets for getting all mark prices

### LINKED ISSUE #118: Switch to Websocket API

ryansonshine:
This change implements the Binance WebSocket client as a quick patch for `get_all_market_tickers` to help with API rate limits since it appears this API is called the most frequently.

I probably won't have time to do any more work on this until next weekend, so if anyone wants to pick it up, I have a list of items I was planning but didn't have time to do.

- [x] Add initial implementation for getting all market tickers via WebSockets
- [ ] Implement error handling on message processing callback
- [ ] Implement re-connection logic for when WebSockets disconnect
- [ ] Look into creating a new class for handling WebSocket connections

Fixes #118

### DIFF
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -22,8 +22,9 @@ class BinanceAPIManager:
         self.db = db
         self.logger = logger
         self.ticker_prices = []
+        self.all_price_socket_conn_key = False
         self.bm = BinanceSocketManager(self.binance_client)
-        self.bm.start_all_mark_price_socket(self._process_all_mark_price_socket)
+        self._start_all_mark_price_socket()
         self.bm.start()
 
     def _process_all_mark_price_socket(self, msg):
@@ -37,6 +38,14 @@ class BinanceAPIManager:
                     self.ticker_prices.append({"symbol": ticker["s"], "price": ticker["p"]})
                 else:
                     self.ticker_prices[existing_idx] = {"symbol": ticker["s"], "price": ticker["p"]}
+
+    def _start_all_mark_price_socket(self):
+        conn = self.bm.start_all_mark_price_socket(self._process_all_mark_price_socket)
+        if conn:
+            self.all_price_socket_conn_key = conn
+        else:
+            self.logger.log("Starting all mark price socket failed, retrying")
+            self.retry(self._start_all_mark_price_socket)
 
     def get_all_market_tickers(self):
         """
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -3,6 +3,7 @@ import time
 
 from binance.client import Client
 from binance.exceptions import BinanceAPIException
+from binance.websockets import BinanceSocketManager
 
 from .config import Config
 from .database import Database
@@ -20,12 +21,26 @@ class BinanceAPIManager:
         )
         self.db = db
         self.logger = logger
+        self.cache = []
+        self.bm = BinanceSocketManager(self.binance_client)
+        self.bm.start_all_mark_price_socket(self._process_all_mark_price_socket)
+        self.bm.start()
+
+    def _process_all_mark_price_socket(self, msg):
+        if msg["data"]:
+            data = msg["data"]
+            for ticker in data:
+                existing_idx = next((i for i, item in enumerate(self.cache) if item["symbol"] == ticker["s"]), None)
+                if existing_idx is None:
+                    self.cache.append({"symbol": ticker["s"], "price": ticker["p"]})
+                else:
+                    self.cache[existing_idx] = {"symbol": ticker["s"], "price": ticker["p"]}
 
     def get_all_market_tickers(self):
         """
         Get ticker price of all coins
         """
-        return self.binance_client.get_all_tickers()
+        return self.cache
 
     def get_market_ticker_price(self, ticker_symbol: str):
         """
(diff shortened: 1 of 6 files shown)


binance_trade_bot/database.py
```diff
@@ -19,7 +19,7 @@ class Database:
     def __init__(self, logger: Logger, config: Config, uri="sqlite:///data/crypto_trading.db"):
         self.logger = logger
         self.config = config
-        self.engine = create_engine(uri)
+        self.engine = create_engine(uri + "?check_same_thread=False")
```
j-waters: https://github.com/edeng23/binance-trade-bot/pull/195/files#diff-0cafd665e7243ffb750d9b2c257d2b3b5673a80451c85d6eacef37e5620f91bcR22

binance_trade_bot/binance_api_manager.py
```diff
@@ -20,12 +21,26 @@ def __init__(self, config: Config, db: Database, logger: Logger):
...
+                    self.cache.append({"symbol": ticker["s"], "price": ticker["p"]})
+                else:
+                    self.cache[existing_idx] = {"symbol": ticker["s"], "price": ticker["p"]}
 
     def get_all_market_tickers(self):
         """
         Get ticker price of all coins
         """
-        return self.binance_client.get_all_tickers()
+        return self.cache
```
j-waters: Maybe a more specific name? `ticker_prices`?

binance_trade_bot/crypto_trading.py
```diff
@@ -16,6 +16,11 @@ def main():
     config = Config()
     db = Database(logger, config)
     manager = BinanceAPIManager(config, db, logger)
+
+    while len(manager.cache) == 0:
```
j-waters: I'd move this within `get_all_market_tickers`

j-waters (changes requested):
This is good, maybe we'll be able to remove the whole `get_all_market_tickers` and `get_market_ticker_price_from_list` and just replace it with a straight function call

binance_trade_bot/binance_api_manager.py
```diff
@@ -20,12 +21,32 @@ def __init__(self, config: Config, db: Database, logger: Logger):
...
+            for ticker in data:
+                existing_idx = next(
+                    (i for i, item in enumerate(self.ticker_prices) if item["symbol"] == ticker["s"]), None
+                )
+                if existing_idx is None:
+                    self.ticker_prices.append({"symbol": ticker["s"], "price": ticker["p"]})
+                else:
+                    self.ticker_prices[existing_idx] = {"symbol": ticker["s"], "price": ticker["p"]}
 
     def get_all_market_tickers(self):
```
Pekafu, line 50: I'm wondering if these tickers could and should be implemented as in https://github.com/edeng23/binance-trade-bot/pull/198 with the cached fees.

binance_trade_bot/binance_api_manager.py
```diff
@@ -20,12 +21,32 @@ def __init__(self, config: Config, db: Database, logger: Logger):
         )
         self.db = db
         self.logger = logger
+        self.ticker_prices = []
+        self.bm = BinanceSocketManager(self.binance_client)
+        self.bm.start_all_mark_price_socket(self._process_all_mark_price_socket)
+        self.bm.start()
```
Pekafu, line 28: After implementing these closing the program should be done as introduced here: https://python-binance.readthedocs.io/en/latest/websockets.html#close-and-exit-program

I've been experimenting with it and it seems we could actually do it here: https://github.com/edeng23/binance-trade-bot/blob/7e1f72c80d1e6ef05249baff865c1df6c2e1a095/binance_trade_bot/crypto_trading.py#L36
some what like this:

    try:
        while True:
            schedule.run_pending()
            time.sleep(1)
        finally:
            manager.bm.close()
            reactor.stop()

binance_trade_bot/binance_api_manager.py
```diff
@@ -20,12 +21,32 @@ def __init__(self, config: Config, db: Database, logger: Logger):
         )
         self.db = db
         self.logger = logger
+        self.ticker_prices = []
+        self.bm = BinanceSocketManager(self.binance_client)
+        self.bm.start_all_mark_price_socket(self._process_all_mark_price_socket)
```
Pekafu: starting socket might be worth doing in it's own function e.g.

    def start_all_price_socket(self):
        conn = self.bm.start_all_mark_price_socket(self._process_all_mark_price_socket)
        if type(conn) == bool and conn == False
             self.logger.log("Starting all mark price socket failed, retrying")
             # call retry here
        else:
            self.all_price_socket_conn_key = conn

binance_trade_bot/binance_api_manager.py
ryansonshine: Did you intend to reference a different PR? I'm not sure I understand.

Pekafu: Fixed

j-waters: If you mean replace `get_all_market_tickers` with calls to `get_market_ticker_price` every time I agree

j-waters:
I continued this PR in https://github.com/edeng23/binance-trade-bot/pull/211, I hope that's okay! Your work was a great foundation!

ryansonshine:
@j-waters Sounds good to me! Thanks

PR #208, opened 2021-03-16
Make fee automation configurable

Pekafu:
Some ppl are experiencing issues with request weight being used with the current fee implementation.

### DIFF
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -6,7 +6,7 @@ from binance.client import Client
 from binance.exceptions import BinanceAPIException
 from cachetools import TTLCache, cached
 
-from .config import Config, TRADE_FEE_OPTION
+from .config import Config
 from .database import Database
 from .logger import Logger
 from .models import Coin
@@ -20,7 +20,11 @@ class BinanceAPIManager:
             config.BINANCE_API_SECRET_KEY,
             tld=config.BINANCE_TLD,
         )
-        self.TRADE_FEE = config.TRADE_FEE
+
+        try:
+            self.TRADE_FEE = float(config.TRADE_FEE)
+        except ValueError:
+            self.TRADE_FEE = None
         self.db = db
         self.logger = logger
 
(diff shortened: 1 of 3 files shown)


Pekafu:
I'll fix linting in minute and will add mention to the readme

j-waters:
I'd say instead of giving the user those 3 options, maybe just give them the option between `auto` and a number. Then in the README we give a link to the default binance fees

binance_trade_bot/binance_api_manager.py
```diff
@@ -32,6 +37,9 @@ def get_using_bnb_for_fees(self):
         return self.binance_client.get_bnb_burn_spot_margin()
 
     def get_fee(self, origin_coin: Coin, target_coin: Coin, selling: bool):
+        if self.TRADE_FEE:
```
j-waters: Why not do 
```suggestion
        if self.config.TRADE_FEE == 'auto':
            ...
        else:
            return float(self.config.TRADE_FEE)
```

We don't have any other configuration in the `__init__` method and I think we should keep it that way

Pekafu: true, yet binance_api_manager doesn't have config itself. Need to use self.binance_client.config.TRADE_FEE

Pekafu: True, yet binance_api_manager doesn't store config

j-waters: I'd make it store the config then. I know another PR has also done that.

Pekafu:
After merging #211 this isn't as nessessary as before

rel-s:
> 
> 
> After merging #211 this isn't as nessessary as before

@Pekafu how so?

Pekafu:
Api call weight won't be exceeded as easily as currently

edeng23:
I think that it's a good idea to have this optionality regardless of #211.

Blaklis:
Considering the US tld seems to return an empty list, this PR would fix a blocker for them, as it currently crashed trying to retrieve the fees for specific pairs.

edeng23:
@Pekafu Can you please rebase?

Pekafu:
@edeng23 rebased

yudataguy:
Hope this get merged soon, implemented the code to get the binance.us `trade_fee` issue sorted out finally

rel-s (changes requested):
Totally forgot about this, sorry...
I think that before merging we should add the new `trade_fee` option to `.user.cfg.example`, as we did with previous config changes. Other than that, nothing to add :)

PR #209, opened 2021-03-16
Add dependabot scanning of updated dependencies

danstis:
As a developer on this project, 
I want dependencies to be automatically updated by GitHubs dependabot, 
so I don't have to follow updates of all of the dependencies.


danstis:
This may require flicking the switch on the repo settings as well @edeng23

j-waters:
Just playing devil's advocate but what if a dependency update breaks something? The reason we started pinning dependencies was to ensure everything was compatible. Does this automatically commit the changes or just create a PR?

danstis:
This will create a dependabot PR, that will rebase automatically on each new update.
As we get tests in play for everything, this will enable the CI process to automatically create the PR and run the tests for us. :)

I should add the PR will include the release notes for each recommended package version, as well as an idea of how many other repos have had issues merging the recommended package versions.

j-waters:
That sounds great!

And yeah even better once we've got some tests written

danstis:
Here is an example, it's an amazing system:
image

PR #220, opened 2021-03-20
Make supported_coins_list an example file

cornytrace:
This will make supported_coins_list an example file similar to .user.cfg.example, to prevent hassle when switching branches or accidental commits of supported_coins_list.


j-waters:
There was a PR ages ago that added the supported coin list to the config file, which I don't think is a bad idea. This is a good idea for now though

cornytrace:
That sounds like a good idea. This is indeed more of a stop-gap measure

rel-s (approved):

PR #211, opened 2021-03-16
Add Binance websocket support

j-waters:
Continuation of https://github.com/edeng23/binance-trade-bot/pull/205, thank you so much to @ryansonshine for the initial implementation!

A cache of the current ticker prices and the user's current balance are now stored in the `BinanceApiManager`. This cache is updated live by two webscoket connections.

Eventually we'll also be able to use this to wait for trades to complete, but I don't want to add anything that will conflict with #195 too much.

In the event of the socket disconnecting, it will attempt to reconnect. I had to extend the binance library a bit to make sure that was working. If it can reconnect, it will clear the caches to make sure new data is fetched, just in case a socket event was missed. If it can't, currently the bot will just continue running without the sockets. Since ticker prices will never update, the bot won't buy anything, so this won't do too much damage. I'll think of how to best handle this, any ideas would be appreciated.

This massively reduces the number of API calls we have to make, and makes scouting so much faster. The `get_fees` method also runs instantly now.

PR #230, opened 2021-03-22
Fix bridge_scout in default strategy

j-waters:
Set current coin after jumping from bridge coin

My bad, I forgot to extend the `bridge_scout` method in the default strategy to update the current coin

Also, with the `get_fee` method, `bridge_scout` is really really slow. I think #211 or #208 should be a priority


cornytrace:
Consider this scenario: the bridge is USDT, and the current coin is 1 ETH. Some time has passed since the buy, so ETH is no longer the preferred trade.
I use fiat currency to buy more USDT, so the bridge balance is above min_notional.
If I understand correctly, the bridge scout will now use the bridge coin to buy another coin, and update the current coin, overwriting the ETH as current coin.
EDIT: also the line `return coin` at line 159 of auto_trader is indented such that it returns a new coin even if the bridge currency amount is below min_notional

j-waters:
@cornytrace ah you're right, thanks. I've changed it so that `bridge_scout` will only run if there isn't enough of the current coin.

PR #232, opened 2021-03-23
Init testing

tamasmagyar:
- Added pytest as a dependency
- Added unit tests for config
- Added unit test job for ci


tests/test_data/user_config.py
```diff
@@ -0,0 +1,17 @@
+API_KEY = 'vmPUZE6mv9SD5VNHk4HlWFsOr6aKE2zvsw0MuIgwCIPy6utIco14y7Ju91duEh8A'
+SECRET_KEY = 'NhqPtmdSJYdKjVHjA7PZj4Mge3R5YNiP1e3UZjInClVN65XAbvqqM6A7H5fATj0j'
```
GrigorovskyA: be careful with your api pair

tamasmagyar: Thanks, that's the default from .user.cfg.example

PR #234, opened 2021-03-23
fix retrieval of BNB fees switch on Binance API

Blaklis:
The `get_using_bnb_for_fees` function will return an object containing e.g. :

```
{
   "spotBNBBurn":true,
   "interestBNBBurn": false
}
```

Used at is, it will evaluate to true each time. Fix get the `spotBNBBurn` property instead.


j-waters:
Can't believe I missed that thanks for catching

edeng23:
Good catch, thanks!

PR #245, opened 2021-03-27
Fix Period Parameter in API Server

jhoogstraat:
This will fix the period api parameter. Before it was using 1d as a fixed value, so you could only get 1 hour, 1 day, etc...


jhoogstraat:
Why is the second check failing?

j-waters:
Good catch, any reason why `.pylintrc` was modified?

As for the second check failing, for some reason it does that sometimes. Try force-pushing something and then removing it again to make the tests run again

jhoogstraat:
The period filter function now has seven possible returns, which pylint does not allow. The change increases the maximum number of return statements to 7.

PR #256, opened 2021-04-07
db: control logging of scout history using a parameter.

pwnfoo:
A lot of people seem to be having issues with the `self.db.log_scout` write overhead.

This PR adds a config using which users can disable it without making code changes. In order to ensure backwards compatibility, the update will also inform users that there's a new param and it needs to be set :)

### DIFF
--- a/.user.cfg.example
+++ b/.user.cfg.example
@@ -8,3 +8,4 @@ hourToKeepScoutHistory=1
 scout_multiplier=5
 scout_sleep_time=5
 strategy=default
+verbose_scout_logging=False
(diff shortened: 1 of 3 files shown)


Pekafu (approved):

pwnfoo:
rebased!

j-waters:
I'll just say I added the logging of scout history when I had lots of free time and was planning on building a web interface, and that obviously hasn't happened! It's currently just used by the API, and I don't know if anyone uses the API or that part of the API. I'd be perfectly happy completely disabling it for now, or merging this PR with the default being off. Then when I or someone else gets round to actually needing it we can re-enable it and try and make it a bit faster :)

idkravitz:
I tried to do batch writing of log_scouts and it was very fruitful, the speed up negated any performance issue, so it may be a better solution, than disable it completely

Issue #260, opened 2021-04-11
KeyError: 'success' when using Binance IP restrictions (and wrong IP)


Ixtalo:
When using Binance.com IP restrictions and running the bot from a wrong IP.

```
Traceback (most recent call last):
  File "binance-trade-bot/binance_trade_bot/scheduler.py", line 25, in _run_job
    super()._run_job(job)
  File "binance-trade-bot/venv/lib/python3.8/site-packages/schedule/__init__.py", line 147, in _run_job
    ret = job.run()
  File "binance-trade-bot/venv/lib/python3.8/site-packages/schedule/__init__.py", line 491, in run
    ret = self.job_func()
  File "binance-trade-bot/binance_trade_bot/strategies/default_strategy.py", line 40, in scout
    self._jump_to_best_coin(current_coin, current_coin_price, all_tickers)
  File "binance-trade-bot/binance_trade_bot/auto_trader.py", line 128, in _jump_to_best_coin
    ratio_dict = self._get_ratios(coin, coin_price, all_tickers)
  File "binance-trade-bot/binance_trade_bot/auto_trader.py", line 115, in _get_ratios
    transaction_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True) + self.manager.get_fee(
  File "binance-trade-bot/binance_trade_bot/binance_api_manager.py", line 43, in get_fee
    base_fee = self.get_trade_fees()[origin_coin + target_coin]
  File "binance-trade-bot/venv/lib/python3.8/site-packages/cachetools/decorators.py", line 22, in wrapper
    v = func(*args, **kwargs)
  File "binance-trade-bot/binance_trade_bot/binance_api_manager.py", line 36, in get_trade_fees
    return {ticker["symbol"]: ticker["taker"] for ticker in self.binance_client.get_trade_fee()["tradeFee"]}
  File "binance-trade-bot/venv/lib/python3.8/site-packages/binance/client.py", line 2169, in get_trade_fee
    if not res['success']:
KeyError: 'success'
```

`res` is indeed `{'msg': 'Invalid API-key, IP, or permissions for action.'}`

Ixtalo:
I just realized that this is not a binance-trade-bot-issue but in underlying python-binance.
The problem has been fixed: https://github.com/sammchardy/python-binance/commit/ff024f09b72b3db38bc62a6b247078609623a062

PR #261, opened 2021-04-11
update to python-binance==0.7.10 (#260)

Ixtalo:
See https://github.com/edeng23/binance-trade-bot/issues/260

One possible fix is to update python-binance.
The problem has been fixed: https://github.com/sammchardy/python-binance/commit/ff024f09b72b3db38bc62a6b247078609623a062


Pekafu:
update .pre-commit-config.yaml aswell, so lint won't fail

rel-s (approved):
Testing this, could you squash to commits into one?

PR #270, opened 2021-04-13
Fix time sync errors

Blaklis:
As we have now issues with this daily, default to 5000ms, and configurable in user.cfg.


rel-s:
I'm experiencing this issue too! However, I think we should further discuss the consequences of this. Why is the recvwindow initially set lower than this? Can something bad happen if we increase it?

Blaklis:
The recvwindow is initially set to 1000ms; the goal is to ensure that, once you sent a request, it has been received by Binance on this time window. As it's a trading system, operations might be critically dependant of this window. In our case, I'm not sure that we really want to enforce such a low window requirement - we're not doing specific operations that really needs seconds precision.

The worst that can happen here is in case of a network lag - your request will be delayed but still accepted by the server.

Made it configurable to ensure people can still reset it to 1000ms if needed.

rel-s:
I don't know @Blaklis - we've already seen in the past that a delay increase of seconds can cause more trades to get stuck in bridge coin because the ask price is already too low... But perhaps it should be configurable, with a sensible default.

PR #273, opened 2021-04-14
bridge_scout not taking into considering market ticker price

iangorse:
bridge_scout would perform a comparison against the current coins balance vs the minimal required.
However, it should have been current balance multiplied by the ticker price.

Without this, the bridge_scout would fail the conditional check and perform a full scout of every coin combination causing significant delays in scouting.


iangorse:
After further review, I no longer think this is an issue. 

PR #195 that was merged into master, removes the call to the bridge_scout in the default_strategy. 
The `def bridge_scout(self):` is now redundant in default_strategy making this PR ineffective.

Blaklis:
After discussions, it would still be useful to have it on the default strategy. Might be used again.

PR #275, opened 2021-04-14
Fix collate coin in case BTC is in the supported coin list

Blaklis:
Currently crash if BTC is in the supported coin list and is used, trying to get BTCBTC symbol.


rel-s (approved):
Seems legit.

PR #276, opened 2021-04-15
Fix price format on request to Binance API

### LINKED ISSUE #281: APIError(code=-1100): Illegal characters found in parameter 'price'
I get 
 APIError(code=-1100): Illegal characters found in parameter 'price' 

image

Hot to fix it

thx!

Blaklis:
Otherwise, it crashes when using something like BTC as a bridge, due to very low prices.


rel-s:
Could you collapse these commits so we don't merge 4 "lint" commits?

Blaklis:
Sure I will!

rel-s:
Btw, I didn't test it but isn't it problematic that `order_quantity` used to be a number, and now you're formatting it into a string and passing it to the same function? Does this work as expected?

Blaklis:
I had no problem during my test - I guess this will be used as string to call the Binance API finally.

I'll do further check, just to ensure it's correct.

idkravitz:
I got a report of
```
APIError(code=-1111): Precision is over the maximum defined for this asset.
```
after applying this fix from one user. Unfortunately he is bad at english, so no useful info extracted from him)

idkravitz:
Well I got some useful info, apparently this happened with SHIB coin as current coin. And it has nothing to do with price format, since it happens even after reverting.

safciplak:
i encountered BTCDOWN. @idkravitz mentioned SHIB. i think if coin price 0.00.... $ problem thrown.

kultawat:
confirm this fixed work for me.

PR #285, opened 2021-04-17
Make the log message output human readable

xioxu:
The original log output is hard to view in linux.


Blaklis:
How hard it is? It only replaces the old message by the new one to avoid flooding the console.

PR #290, opened 2021-04-21
Reduce amount of `binance_client.get_account()` calls.

nazegnl:
Fixes a couple of locations where there were redundant calls to the `self.binance_client.get_account()["balances"]`


rel-s (approved):
Seems like a good idea, will test this soon.

rel-s:
Works on my end, we'll move forward with merging this after #328 is merged.

levdavid:
#328 was merged, should we merge it in?

PR #293, opened 2021-04-21
Reduce multiple coin strategy excessive logging.

nazegnl:
Implement the same logic as the default strategy for displaying the scout message.

### DIFF
--- a/binance_trade_bot/strategies/multiple_coins_strategy.py
+++ b/binance_trade_bot/strategies/multiple_coins_strategy.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from binance_trade_bot.auto_trader import AutoTrader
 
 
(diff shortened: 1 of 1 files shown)


binance_trade_bot/strategies/multiple_coins_strategy.py
```diff
@@ -33,7 +35,11 @@ def scout(self):
 
             # Display on the console, the current coin+Bridge, so users can see *some* activity and not think the bot
             # has stopped. Not logging though to reduce log size.
-            self.logger.info(f"Scouting for best trades. Current ticker: {coin + self.config.BRIDGE} ", False)
+            print(
```
erenatas, line 38: Why are you using print? How is this going to help with reduce the logging?

nazegnl: This way the logs are a) not written to file, and b) the `end="\r"` will make it so that it only prints one line in stdout. 
So in my opinion this reduces the "log spam" drastically.

erenatas: IMO scouting logs are also valuable in terms of IO and networking problems.

nazegnl: I think there should be specific logging for networking or other issues rather then and relying on a other message.

Blaklis: I'm ok with @nazegnl, and it's also how it's done in default strategy. Should be less spammy then

erenatas: > I think there should be specific logging for networking or other issues rather then and relying on a other message.

An example to this was, I started using the bot from a USB 3.0 stick. After a day, it turns out that the write speed on the stick got so low, it was not able to do scouting. The only way to understand it was from the time difference on each scout logs. There was no errors because it was not an error essentially. By not using the logger, we not only going to lose the scouting logs, also will lose the time stamps. Therefore I do not like it. Not using a logger is a bad practice.

PR #295, opened 2021-04-22
Revert changes on strategies

Blaklis:
I don't think the current version is correct : 

- The multiple coin strategy should never take care of the current_coin value - it should scout for each assets we own, which is not the case anymore I think?
- Both strategies should still scout with the bridge if ever we have no coins, allow to recover from manual failure (cancelling a trade manually, cutting the bot in the middle of a trade, etc).

### DIFF
--- a/binance_trade_bot/strategies/default_strategy.py
+++ b/binance_trade_bot/strategies/default_strategy.py
@@ -3,12 +3,18 @@ import sys
 from datetime import datetime
 
 from binance_trade_bot.auto_trader import AutoTrader
+from binance_trade_bot.models import Pair
 
 
 class Strategy(AutoTrader):
     def initialize(self):
         super().initialize()
         self.initialize_current_coin()
+
+    def transaction_through_bridge(self, pair: Pair, all_tickers):
+        super().transaction_through_bridge(pair, all_tickers)
+
+        self.db.set_current_coin(pair.to_coin)
 
     def scout(self):
         """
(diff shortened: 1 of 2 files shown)


iangorse:
I believe there is a bug in the bridge_scout method of the default strategy.

```
def bridge_scout(self):
        current_coin = self.db.get_current_coin()
        if self.manager.get_currency_balance(current_coin.symbol) > self.manager.get_min_notional(
            current_coin.symbol, self.config.BRIDGE.symbol
        ):
            # Only scout if we don't have enough of the current coin
            return
        new_coin = super().bridge_scout()
        if new_coin is not None:
            self.db.set_current_coin(new_coin)
```

An example I had was that I had 4.4 QTUM. 
get_currency_balance(current_coin.symbol) returned 4.4 (of QTUM)
get_min_notional(current_coin.symbol, self.config.BRIDGE.symbol) returned 10.0 (of USDT)

This conditional check would fail.
In reality 4.4 QTUM was worth over 100 USDT at the time so should have returned before it started to do the super().bridge_scout

My PR #273  handled this.

dv-hoang:
@Blaklis the check current coin in multiple coin strategy to make sure the bot buys another coin at the correct rate. For example, the bot sold coin A but time out in buy coin B, it should buy another coin base on the rate of coin A instead of buying the random coin.

halilural (approved):

Blaklis:
@kydrenw it's a little tricky and I guess this needs to be managed in another way.

idkravitz:
Keeping transactions (pending jumps) in db might do the thing and it will also allow us to restart bot in the middle of jump

j-waters:
Yeah the database does keep track of the status of transactions but nothing is really done with it, I'm sure we could use that instead.

PR #298, opened 2021-04-25
Check order_quantity before creating buy order

dv-hoang:
This PR to resolve 2 issues when recovering the scout in the multiple coin strategy:
- If sold coin and timeout in a buy order, the bot does not update the current coin to calculate the rate base on the sold coin when recovering 
- If the user stops the bot, manual cancel the buy order then start the bot, the bot will keep buying the other coin with quality is zero when reaching the rate.


werim:
Writing the same codes a min ago and saw this...

werim (approved):

PR #300, opened 2021-04-25
Fixed display error of the yield with percentage format

ad19900913:
as title


Blaklis:
This is a duplicate of #291 and should be closed as is.

rel-s:
Yep, sorry @ad19900913! We appreciate the contribution.

PR #303, opened 2021-04-27
Update multiple_coins_strategy.py

werim:
Solution to type error for multiple_coins_strategy.

### DIFF
--- a/binance_trade_bot/binance_api_manager.py
+++ b/binance_trade_bot/binance_api_manager.py
@@ -68,6 +68,10 @@ class BinanceAPIManager:
         Get ticker price of all coins
         """
         return AllTickers(self.binance_client.get_all_tickers())
+#seco 27-04-21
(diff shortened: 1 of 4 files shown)


Blaklis:
This is not how it should be patched. The scouting message should only dispaly coin instead of current_coin.

rel-s:
Hi, please update this PR with some more information. An informative title, some text detailing what you fixed - perhaps even linking to a related issue if there is one? thanks!

Blaklis:
This PR intended to patch the error on multiple_coins strategy with `current_coin` being printed, which sometimes leads to a `NoneType` being concatenated to a string.

This PR now have multiple commits that is not related to the initial fix that was provided - surely because it's a PR from the author master branch. The fix was also incorrect, so I guess this can be closed now.

rel-s:
Yep, closing.

PR #310, opened 2021-04-28
WIP: Elon bot

jamielaff:
## WIP
- [x] Reimplement class from existing repo
- [x] Remove BRIDGE usage, run with current_coin
- [x] Move into its own strategy
- [ ] Hook up properly with DB
- [ ] Unit tests
- [ ] Personal tests


erenatas:
When manipulation becomes a reality...

Blaklis:
Would be better if it's implemented as a new strategy instead ?

jamielaff:
> Would be better if it's implemented as a new strategy instead ?

Could be - it'll probably be a very long time between executions though 😄 
I'm also not sure on the behaviour if you have this + default/multi strategy running. Thoughts?

geeks121:
this bot should be added for strategy

rel-s:
This definitely won't be merged as a default behaviour. Perhaps as another strategy, although this deviates pretty far from the bot's original purpose...

PR #314, opened 2021-05-02
Do not jump if current_coin_price is below bought price…

werim:
Do not jump if current_coin_price is below bought price in default_strategy.py . Take the price used to bought current_coin, from trade_history table. Function get_price_old in Database used to get any coin's bought price. You may set a stoploss or take profit.

### DIFF
--- a/binance_trade_bot/database.py
+++ b/binance_trade_bot/database.py
@@ -115,6 +115,17 @@ class Database:
             session.expunge(coin)
             return coin
 
+    def get_price_old(self, alt_coin_id: Union[Coin, str]):
+        session: Session
+        with self.db_session() as session:
+            trade: Trade = session.query(Trade).filter(Trade.alt_coin_id == alt_coin_id.symbol,
+                                                       Trade.state == 'COMPLETE'). \
+                order_by(Trade.datetime.desc()).first()
+            session.expunge(trade)
+            if trade.selling: return
+            price: float = trade.crypto_trade_amount / trade.alt_trade_amount
+            return price
+
     def get_pair(self, from_coin: Union[Coin, str], to_coin: Union[Coin, str]):
         from_coin = self.get_coin(from_coin)
         to_coin = self.get_coin(to_coin)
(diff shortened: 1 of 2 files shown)


eltjor:
I think this is a good impovement, however I suggest extend this with a fallback to the latest buy order for this currency with API call client.get_all_orders(symbol={current_coin + self.config.BRIDGE}, limit=10) and look for last filled order. In case there is no data available in the database.

GrigorovskyA:
What's the point? 

* Let's imagine we are in coin A - which fell 5%. There is also coin B which fell by 10%.
It is advantageous for us to take coin B at a given time. And according to your logic, this will not happen and the bot will not be able to increase the amount of currency.

* And if bot successfully tooks a coin at the top of the graph - and the graph began to fall - you will never get out of the coin until it rises again.

werim:
You can set the strategy for stop-loss or take-profit. I just draw the frame.

Blaklis:
Not a good idea, in my opinion. The bot is focused on increasing the number of coins, and don't care about fiat. Seems clearly against this very main goal.

werim:
> Not a good idea, in my opinion. The bot is focused on increasing the number of coins, and don't care about fiat. Seems clearly against this very main goal.

Finds very low coins, but sells before coin pumps. When I figured that, think it needs to wait to pump.

MohammedDarwish-0:
> You can set the strategy for stop-loss or take-profit. I just draw the frame.

Could you pass the stop-lose strategy?

werim:
> > You can set the strategy for stop-loss or take-profit. I just draw the frame.
> 
> Could you pass the stop-lose strategy?

sth like this:
        if old_price is not None and old_price * (1-ratio) < current_coin_price < old_price * (1 + ratio):
            old = "{:.6f}".format(old_price)
            current = "{:.6f}".format(current_coin_price)
            print(f"{datetime.now()} - Current price {current} lower than bought price {old}")
            time.sleep(20)
            return

rel-s:
This definitely won't get merged to the default strategy. As @Blaklis noted, this goes against the very purpose of this bot - in my opinion, even as an optional strategy it doesn't fit in this bot's vision. We do not focus on fiat, but rather on crypto currency increases. I'm closing this PR, if you want to reopen it as a different strategy and have this discussion over there feel free.

PR #317, opened 2021-05-04
WIP: Multiple concurrent active coins

laffy:
**WORK IN PROGRESS - DO NOT MERGE**

The version will split your USDT balance and trade on multiple coins at any one time.

It will not allow you to trade 1 coin into another active coin, thereby distributing your trading and risk. It will however trade into another non-active coin as long as the ratio is above 0, as per all the other strats.

I am running this on production now, I will let you know how it works out.

Also merged [IDK master](https://github.com/idkravitz/binance-trade-bot/commit/3e722a30d233845096b6bb77171e834401d22af5) (webhooks) in on 16th May /cc @idkravitz 

There is also a `sell coin for profit` strategy - more to follow.


binance_trade_bot/database.py
```diff
@@ -145,6 +145,31 @@ def get_pairs(self, only_enabled=True) -> List[Pair]:
...
+            return coins
+
+    def set_coin_to_active(self, bought_coin):
+        session: Session
+        with self.db_session() as session:
+            bought_coin.active = True
+            session.add(bought_coin)
+            session.flush()
+            self.logger.info(f"--- Activating coin: {bought_coin}")
+            self.send_update(bought_coin)
```
laffy, line 165: For some reason this creates an error, likely from flushing the session, however it does not prevent the bot from continuing to run.

I could not find another way to update active coins within the bot thread, perhaps somebody who knows more about Python than me can help?

image

.user.cfg.example
```diff
@@ -1,6 +1,6 @@
 [binance_user_config]
-api_key=vmPUZE6mv9SD5VNHk4HlWFsOr6aKE2zvsw0MuIgwCIPy6utIco14y7Ju91duEh8A
-api_secret_key=NhqPtmdSJYdKjVHjA7PZj4Mge3R5YNiP1e3UZjInClVN65XAbvqqM6A7H5fATj0j
+api_key=XXX
```
j-waters, line 2: That API key is the example they use on the API documentation no need to change it

binance_trade_bot/auto_trader.py
```diff
@@ -40,6 +42,7 @@ def transaction_through_bridge(self, pair: Pair, all_tickers: AllTickers):
         result = self.manager.buy_alt(pair.to_coin, self.config.BRIDGE, all_tickers)
 
         if result is not None:
+            # TODO: Do I need to change this?
             self.db.set_current_coin(pair.to_coin)
```
j-waters, line 45: There's some debate about whether this will remain, I don't think you need to worry - it might be removed at some point (in my opinion this class shouldn't have any references to 'current_coin')

binance_trade_bot/auto_trader.py
```diff
@@ -142,11 +145,21 @@ def _jump_to_best_coin(self, coin: Coin, coin_price: float, all_tickers: AllTick
...
 
+        can_trade_this_coin = True
+
         # if we have any viable options, pick the one with the biggest ratio
         if ratio_dict:
             best_pair = max(ratio_dict, key=ratio_dict.get)
-            self.logger.info(f"Will be jumping from {coin} to {best_pair.to_coin_id}")
-            self.transaction_through_bridge(best_pair, all_tickers)
+            # Do not allow us to trade with another active coin
+            for active_coin in self.db.get_active_coins():
```
j-waters: Instead I would add an optional `excluded_coins` argument to some of these methods - ideally, the `AutoTrader` class shouldn't have stuff that is only really used by one strategy

binance_trade_bot/auto_trader.py
```diff
@@ -142,11 +145,21 @@ def _jump_to_best_coin(self, coin: Coin, coin_price: float, all_tickers: AllTick
...
+                if active_coin.symbol == best_pair.to_coin_id:
+                    can_trade_this_coin = False
+
+            if can_trade_this_coin:
+                self.logger.info(f"Will be jumping from {coin} to {best_pair.to_coin_id}")
+                self.transaction_through_bridge(best_pair, all_tickers)
+            else:
+                self.logger.info(f"--- Skipping trade for {coin}... new coin {best_pair.to_coin_id} is already active")
 
     def bridge_scout(self):
```
j-waters:
```suggestion
    def bridge_scout(self, excluded_coins=None):
        if excluded_coins is None:
            excluded_coins = []
```

See my previous comment

binance_trade_bot/database.py
```diff
@@ -257,7 +282,7 @@ def migrate_old_state(self):
             os.rename(".current_coin_table", ".current_coin_table.old")
             self.logger.info(".current_coin_table renamed to .current_coin_table.old - " "You can now delete this file")
 
-
+# TODO: Can I move this out to a model class?
```
j-waters: I don't think so since this isn't a database model per se - the actually model is `Trade`, this is just a wrapper that facilitates changing the actual database model

binance_trade_bot/strategies/multiple_active_coins_strategy.py
```diff
@@ -0,0 +1,44 @@
+from datetime import datetime
```
j-waters, line 1: If you wanted to add this to the existing multiple_coins_strategy I don't think that would be an issue

j-waters (changes requested):
This was one of the main reasons I created the multiple coins strategy in the first place, so thank you so much for putting this together!

A couple of thoughts (feel free to disagree):
* I think this should be generalised to allow for more than 2 coins to be active, is there a reason why it's limited to 2?
* I'm pretty cautious about adding a new row to the database, although an 'active' column is a fairly good idea and could even replace the `current_coin` table as a more robust alternative (especially as more strategies are added), would have to get @edeng23's opinion on that. The real issue is you need to have a system for migration - if your PR was merged, people would have to delete their database to use the new version, which obviously isn't ideal. Now, a migration system is way beyond the scope of this PR, so I don't suggest you add one, it just means that as it stands this is unlikely to be merged. Would it be possible to simply infer which coins are active by which coins the user owns? So if I have ETH and XLM and nothing else, we assume those are my 'active coins'? It might not be ideal but it would mean this could be merged
* There has been some discussion previously about configuring strategies, and that we should have different sections in the config file for different strategies, or separate config files entirely. Not sure what the best solution is, this is just the first PR to add a new strategy so we might want to  discuss it

Sorry if some of that came across as negative btw, I really like the PR and my comments are just to increase the chances of it being merged

binance_trade_bot/database.py
j-waters: Also that's out of the scope of this PR

laffy:
Hi @j-waters thanks for your comments.

I should have made this a draft PR that is my mistake sorry, it's definitely not ready to be merged yet as I know I haven't separated out the concerns fully yet.

I will full review your feedback this evening.

binance_trade_bot/strategies/multiple_active_coins_strategy.py
laffy: Ah yes, I needed this as my `production` version runs on Postgres not SQLite.

laffy: It already is: 

https://github.com/edeng23/binance-trade-bot/blob/master/binance_trade_bot/strategies/multiple_coins_strategy.py#L1

j-waters: Oh weird sorry on Github it looks like this is a brand new file rather than a modified one

.user.cfg.example
laffy: I think they should be removed from the example file completely to prevent somebody accidentally sending their real API keys to somebody else.

laffy:
@j-waters I have addressed your first point. It was set to 2 mostly to confirm it worked. I have now made this a variable and tested it, so this can be set to any number you want now.

I have also addressed your concern with strategy specific code in the auto trader class.

Nesonica:
Hi there, thanks for the commit! How can I pull this into my current Pycharm Project? :-)

.user.cfg.example
erenatas: User should never send someone his/her user.cfg directly. That's user's fault and not related to this PR.

binance_trade_bot/database.py
laffy: Can anybody help with this error?

.user.cfg.example
laffy: That is true, but why even risk it? :)

README.md
```diff
@@ -70,11 +73,13 @@ BRIDGE_SYMBOL: USDT
...
 API_SECRET_KEY: NhqPtmdSJYdKjVHjA7PZj4Mge3R5YNiP1e3UZjInClVN65XAbvqqM6A7H5fATj0j
 SCOUT_MULTIPLIER: 5
-SCOUT_SLEEP_TIME: 5
+SCOUT_SLEEP_TIME: 1
 TLD: com
 STRATEGY: default
 BUY_TIMEOUT: 0
 SELL_TIMEOUT: 0
+BUY_ORDER_TYPE: limit
+SELL_ORDER_TYPE: market
```
laffy, line 82: TODO: Add to readme.

laffy:
Too out of sync now. Not going to work on this anymore.

556isback:
@laffy
Hi
I am wondering if I could have your permission to implement your code on #361, It could really help with the conflict between my modification and  multiple coin strategy.

PR #346, opened 2021-05-22
Correcting the trading fees for the backtest

AMilkov:
The fees for the backtest are 10 times bigger than the real ones
https://www.binance.com/en/fee/schedule


MasaiasuOse:
Isn't it better to set the fee to the dafault one and not to the BNB trading fee (for VIP 0)?

AMilkov:
> Isn't it better to set the fee to the dafault one and not to the BNB trading fee (for VIP 0)?

@MasaiasuOse,
I don't think so - in real life, everybody is using BNB for the fees. 
It is safe to assume that reduced fees are more probable than the default ones.

PR #352, opened 2021-05-27
Backtesting Speedup: workaround for python-binance==0.7.11 bug that means price fetch limits are ignored

Knifa:
We call `get_historical_klines` to fetch old ticker prices in the back testing module, which involves passing in a target symbol, etc. but more importantly a begin date, an end date, and a count limit. 

Unfortunately, the library just ignores this.

https://github.com/sammchardy/python-binance/blob/0.7.11/binance/client.py#L873

```python
def get_historical_klines(self, symbol, interval, start_str, end_str=None, limit=500, spot=True)
    # ...
    return self._historical_klines(symbol, interval, start_str, end_str=None, limit=500, spot=True)
```

Fortunately, we can just call `_historical_klines` ourselves directly to workaround the problem. This gives a huge speed boost when fetching.

### DIFF
--- a/binance_trade_bot/backtest.py
+++ b/binance_trade_bot/backtest.py
@@ -61,7 +61,10 @@ class MockBinanceManager(BinanceAPIManager):
                 end_date = datetime.now()
             end_date = end_date.strftime("%d %b %Y %H:%M:%S")
             self.logger.info(f"Fetching prices for {ticker_symbol} between {self.datetime} and {end_date}")
-            for result in self.binance_client.get_historical_klines(
+
+            # _historical_klines must be used as get_historical_klines has
+            # bugged args (ignores end_date and limit) on python-binance==0.7.11
+            for result in self.binance_client._historical_klines(  # pylint: disable=no-member, protected-access
                 ticker_symbol, "1m", target_date, end_date, limit=1000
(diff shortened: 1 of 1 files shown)


Knifa:
Hmm, just realised the main branch is using `0.7.10` which doesn't have this problem. I'm running off something with the Unicorn Websockets base which bumps it up, so maybe this needs to wait.

edeng23:
Noted. Thanks!

Knifa:
This'll need merged now, @edeng23 👍

idkravitz:
Maybe we may just use fresh python-binance? They fixed it somewhere around 1.0.8 version. The additional benefit is that they switched everything from wapi to sapi and binance going to disable wapi soon. If I recall correctly bot uses wapi when getting fees.

Knifa:
> Maybe we may just use fresh python-binance? They fixed it somewhere around 1.0.8 version. The additional benefit is that they switched everything from wapi to sapi and binance going to disable wapi soon. If I recall correctly bot uses wapi when getting fees.

That sounds ideal. :smile: This PR is definitely just a hack to get things going faster!

Knifa:
Superseded by #352 :+1:

PR #361, opened 2021-06-05
Sell percentage & buy percentage

556isback:
Sell percentage
Set the percentage to 80 means while the bot sells the current coin it leaves a small portion of the current coin, prevent the fact that it might moves up rocket high and the bot never purchase this coin again.
Buy  percentage
Set the percentage to 80 means while the bot buys the current coin it leaves a small portion of the bridge coin  when the market is on the uptrend or extra  bridge coin if the account have that extra when the market is on the downtrend  , which means it take profit on the way up and buy more on the way down.


556isback:
However this modification has conflict with the multiple coin strategy, it keeps selling the original coin until it run out of balance

PR #371, opened 2021-06-13
Massively speed-up/reduce memory usage on prune_value_history()

Knifa:
This speeds up `prune_value_history()` massively by using SQL statements directly, rather than fetching them all and then updating.

As it stands, SQLAlchemy will always fetch all of the results at once which takes up a ton of time and memory, especially as you go to loop through them. This just uses UPDATE straight up, with a subquery that grabs IDs based on the GROUP BY.

I don't have real benchmarks but we're talking going from 30-45s and causing regular OOMs on my containers to taking 5-10s and using little to no memory, from just testing myself.

### DIFF
--- a/binance_trade_bot/database.py
+++ b/binance_trade_bot/database.py
@@ -7,7 +7,7 @@ from typing import List, Optional, Union
 
 from socketio import Client
 from socketio.exceptions import ConnectionError as SocketIOConnectionError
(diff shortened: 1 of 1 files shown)


Knifa:
Note: this needs SQLAlchemy 1.4, oops!

PR #372, opened 2021-06-13
Upgrade python-binance to 1.0.12 and pylint to 2.8.3

idkravitz:
Fresh python-binance fixed backtest slowdown issues, that were introduced in 0.7.11. They also switched from WAPI to SAPI, because WAPI is going to be disabled by binance soon (versions prior to 1.0.x used WAPI, for instance, to obtain trade fees). And they add async support that we may utilize later.

Pylint upgrade is necessary because 2.4.4 version has bugs with handling parametrized type hints like `typing.Optional`. Also https://github.com/pre-commit/mirrors-pylint is marked by author as deprecated, use direct pylint repo instead.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,8 +55,8 @@ repos:
     files: ^(dev-)?requirements\.txt$
     args: [requirements.txt, dev-requirements.txt]
 
-- repo: https://github.com/pre-commit/mirrors-pylint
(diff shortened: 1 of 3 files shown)


idkravitz:
This should resolve issue addressed in https://github.com/edeng23/binance-trade-bot/pull/352

edeng23:
Hey, can you please rebase the PR? Thanks :)

PR #376, opened 2021-06-13
Upgrade sqlalchemy to 1.4.15

Knifa:
Needed for #371.

Mainly 1.4 lets use use 2.0 style statements, where you can create expressions and statements without being tied to a session. Much more flexible.


binance_trade_bot/models/pair.py
```diff
@@ -22,6 +22,7 @@ class Pair(Base):
         select([func.count(Coin.symbol) == 2])
         .where(or_(Coin.symbol == from_coin_id, Coin.symbol == to_coin_id))
         .where(Coin.enabled.is_(True))
+        .scalar_subquery()
```
Knifa, line 25: Fixes warning about implicit subquery with the upgrade.

PR #381, opened 2021-06-21
Update inverse pair ratio when updating thresholds

CryptoIsGarbage:
I think there may be a bug when we are updating the thresholds that we also need to update the inverse pair.
**I'm not too familiar with the ratio math so I'm keen to get some feedback on this and hope someone can prove me wrong :)**

With the code in master, today, when we move from coin FOO -> BAR we will go through the db and update the ratios for all pairs from * to BAR.
When scouting, ratios queried are from BAR to *, so when evaluating if to jump back to FOO an old ratio is used. I believe this is why sometimes the bot will immediately jump back at a loss sometimes.

With this change, the ratio for BAR -> FOO will also be updated, and will ensure that the ratio improves before jumping back to the previous coin. The seems to have improved the bot performance for me over the past few days.

Something I'm not sure about is if we should only update this ratio if the "new" ratio is greater than the old one.

### DIFF
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -42,16 +42,18 @@ class AutoTrader:
         result = self.manager.buy_alt(pair.to_coin, self.config.BRIDGE)
         if result is not None:
             self.db.set_current_coin(pair.to_coin)
-            self.update_trade_threshold(pair.to_coin, result.price)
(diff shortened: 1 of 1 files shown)


idkravitz:
Basically it is same as reset ratios for current coin after each trade, so you actually don't utilize any history of ratios. 
If set max of current and previous it may be better, but will halt bot quickly enough.

Issue #235, opened 2021-03-23
crypto_trading_logger - ERROR - Error while scouting...ent coin


ABTDoug:
Using the US version, on the current version as of this issue report.

Traceback (most recent call last):
  File "/mnt/USBDrive/Python/binance-trade-bot/binance-trade-bot/binance_trade_bot/scheduler.py", line 25, in _run_job
    super()._run_job(job)
  File "/mnt/USBDrive/Python/binance-trade-bot/lib/python3.7/site-packages/schedule/__init__.py", line 147, in _run_job
    ret = job.run()
  File "/mnt/USBDrive/Python/binance-trade-bot/lib/python3.7/site-packages/schedule/__init__.py", line 491, in run
    ret = self.job_func()
  File "/mnt/USBDrive/Python/binance-trade-bot/binance-trade-bot/binance_trade_bot/strategies/default_strategy.py", line 41, in scout
    self._jump_to_best_coin(current_coin, current_coin_price, all_tickers)
  File "/mnt/USBDrive/Python/binance-trade-bot/binance-trade-bot/binance_trade_bot/auto_trader.py", line 129, in _jump_to_best_coin
    ratio_dict = self._get_ratios(coin, coin_price, all_tickers)
  File "/mnt/USBDrive/Python/binance-trade-bot/binance-trade-bot/binance_trade_bot/auto_trader.py", line 116, in _get_ratios
    transaction_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True) + self.manager.get_fee(
  File "/mnt/USBDrive/Python/binance-trade-bot/binance-trade-bot/binance_trade_bot/binance_api_manager.py", line 35, in get_fee
    base_fee = self.get_trade_fees()[origin_coin + target_coin]
KeyError: 'DOGEUSDT'

uhaldejm:
I have a similar issue.
On the first run it fails on the first purchase:

> 2021-03-24 00:36:08,078 - crypto_trading_logger - INFO - Setting initial coin to TRX
> 2021-03-24 00:36:10,432 - crypto_trading_logger - INFO - Purchasing <TRX> to begin trading
> 2021-03-24 00:36:17,142 - crypto_trading_logger - INFO - Failed to Buy/Sell. Trying Again.
> 2021-03-24 00:36:17,148 - crypto_trading_logger - INFO - unsupported operand type(s) for *: 'NoneType' and 'int'
[...]

PR #397, opened 2021-07-31
Fix 235: KeyError: 'code' when calling US API

michaelmohamed:
https://github.com/edeng23/binance-trade-bot/issues/235

### DIFF
--- a/binance_trade_bot/auto_trader.py
+++ b/binance_trade_bot/auto_trader.py
@@ -123,10 +123,16 @@ class AutoTrader:
             # Obtain (current coin)/(optional coin)
             coin_opt_coin_ratio = coin_price / optional_coin_price
 
-            transaction_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True) + self.manager.get_fee(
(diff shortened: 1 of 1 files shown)


binance_trade_bot/auto_trader.py
```diff
@@ -123,10 +123,16 @@ def _get_ratios(self, coin: Coin, coin_price):
...
-                pair.to_coin, self.config.BRIDGE, False
-            )
 
+            transaction_fee = 0.001
+            
+            try:
+                transaction_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True) + self.manager.get_fee(
+                    pair.to_coin, self.config.BRIDGE, False
+                )
+            catch:
```
Kvrnn, line 133: Replace the `catch` with `except` since catch is not syntactically correct.

Kvrnn (changes requested):
This worked perfectly with tld=usd to fix the KeyError

Blaklis:
It should return 0.002 by default, not 0.001.

PR #417, opened 2021-09-08
Margin

bamooxa:
Better core formula.
I've been running for some months now and I haven't come across the instant/unnecessary trade bug (A>B>C instantly).
Every trade results in atleast my scout margin percentage coin profit.


edeng23:
I like this change, it makes more sense than my original formula. Will run a few tests and hopefully merge soon.

binance_trade_bot/auto_trader.py
```diff
@@ -123,13 +123,19 @@ def _get_ratios(self, coin: Coin, coin_price):
...
-                coin_opt_coin_ratio - transaction_fee * self.config.SCOUT_MULTIPLIER * coin_opt_coin_ratio
-            ) - pair.ratio
+            # Fees
+            from_fee = self.manager.get_fee(pair.from_coin, self.config.BRIDGE, True)
+            to_fee = self.manager.get_fee(pair.to_coin, self.config.BRIDGE, False)
+            transaction_fee = from_fee + to_fee - from_fee * to_fee
+
+            if self.config.USE_MARGIN == "yes":
+                ratio_dict[pair] = (
+                    (1- transaction_fee) * coin_opt_coin_ratio / pair.ratio - 1 - self.CONFIG.SCOUT_MARGIN / 100
```
yigitakbulut: Here it should be self.**config**.SCOUT_MARGIN

binance_trade_bot/config.py
```diff
@@ -70,3 +72,6 @@ def __init__(self):
 
         self.SELL_TIMEOUT = os.environ.get("SELL_TIMEOUT") or config.get(USER_CFG_SECTION, "sell_timeout")
         self.BUY_TIMEOUT = os.environ.get("BUY_TIMEOUT") or config.get(USER_CFG_SECTION, "buy_timeout")
+
+        self.USE_MARGIN = os.environ.get("USE_MARGIN") or config.get(USER_CFG_SECTION, "use_margin")
+        self.SCOUT_MARGIN = os.environ.get("SCOUT_MARGIN") or config.get(USER_CFG_SECTION, "scout_margin")
```
yigitakbulut: self.SCOUT_MARGIN = **float(** os.environ.get("SCOUT_MARGIN") or config.get(USER_CFG_SECTION, "scout_margin")**)**

yigitakbulut (changes requested):
Very good! Thank you for this, I think the strategy works better with this.
There is a typo in auto_trader.py, also you should convert the scout margin to float when you import it from config.

yigitakbulut (approved):

lisabbasil:
I'm a bit confused about the fees. They are given as a percentage, so should be multiplied with the volume that will be traded, not with a ratio, or a ratio of ratios. Guess I just don't understand the formula?

rogerthomas84:
I really like this proposed change 👍

muhammadnumair:
Thanks for the change, @edeng23 were you able to test it?

Issue #489, opened 2022-04-02
ImportError: cannot import name 'escape' from 'jinja2'


sssemil:
If you try running `python backtest.py`, you will get this error:
```
Traceback (most recent call last):
  File "/home/emil/Projects/binance-trade-bot/backtest.py", line 3, in <module>
    from binance_trade_bot import backtest
  File "/home/emil/Projects/binance-trade-bot/binance_trade_bot/__init__.py", line 1, in <module>
    from .backtest import backtest
  File "/home/emil/Projects/binance-trade-bot/binance_trade_bot/backtest.py", line 8, in <module>
    from .binance_api_manager import BinanceAPIManager
  File "/home/emil/Projects/binance-trade-bot/binance_trade_bot/binance_api_manager.py", line 10, in <module>
    from .binance_stream_manager import BinanceCache, BinanceOrder, BinanceStreamManager, OrderGuard
  File "/home/emil/Projects/binance-trade-bot/binance_trade_bot/binance_stream_manager.py", line 9, in <module>
    from unicorn_binance_websocket_api import BinanceWebSocketApiManager
  File "/home/emil/Projects/binance-trade-bot/venv/lib/python3.10/site-packages/unicorn_binance_websocket_api/__init__.py", line 3, in <module>
    from unicorn_binance_websocket_api.manager import BinanceWebSocketApiManager
  File "/home/emil/Projects/binance-trade-bot/venv/lib/python3.10/site-packages/unicorn_binance_websocket_api/manager.py", line 40, in <module>
    from unicorn_binance_websocket_api.restserver import BinanceWebSocketApiRestServer
  File "/home/emil/Projects/binance-trade-bot/venv/lib/python3.10/site-packages/unicorn_binance_websocket_api/restserver.py", line 36, in <module>
    from flask_restful import Resource
  File "/home/emil/Projects/binance-trade-bot/venv/lib/python3.10/site-packages/flask_restful/__init__.py", line 3, in <module>
    from flask import request, url_for, current_app
  File "/home/emil/Projects/binance-trade-bot/venv/lib/python3.10/site-packages/flask/__init__.py", line 14, in <module>
    from jinja2 import escape
ImportError: cannot import name 'escape' from 'jinja2' (/home/emil/Projects/binance-trade-bot/venv/lib/python3.10/site-packages/jinja2/__init__.py)
```

Similar issue in another project - https://github.com/pallets/flask/issues/4494

sssemil:
[...]

PR #494, opened 2022-04-17
updating dependencies (see issue #489)

ekurer:


edeng23:
First of all, the build does not work because you didn't update the pre-commit-config YAML file. Did you follow the "Developing" instructions? 
Besides, is bumping Flask not enough? Is introducing two new requirements necessary?

